PR #2, opened 2015-11-24
Alphaharvest

bittirousku:
Alphaharvest is working now, tests also.
Baseharvest still has problems, old and new:
* Testing doesn't work, because parse_node() is not returning items! Is this
    the wrong way to go? How else can I iterate through all the records? Or should
    the testing be done differently?
* It's consistently outputting only 184 records when using a test file of 1000 records! (tests/responses/base/test_1000.xml)
   * It works when using smaller test files.
* Australian thesis server sometimes returns errors: `twisted.python.failure.Failure twisted.internet.error.ConnectionDone: Connection was closed cleanly`.
* Other error messages encountered:
   * `IOError: [Errno socket error] [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)`
   * `IOError: [Errno socket error] EOF occurred in violation of protocol (_ssl.c:590)`
   * `IOError: [Errno socket error] [Errno 104] Connection reset by peer`
* CALtech thesis server was asking for a password
* Namespaces are ignored with brute force. There are better ways?
* JSON pipeline is not writing commas between records right now. The original solution to add ", \n" also added a comma after the last record (not cool).
* Something else?


.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
+
+This directory caches those eggs to prevent repeated downloads.
+
+However, it is safe to delete this directory.
+
```
jacquerie: In fact, this should probably be added to `.gitignore`.

jalavik:
@bittirousku Figured out the main issue we had today with harvesting in the base spider, it was relative vs. absolute xpath expressions: http://doc.scrapy.org/en/latest/topics/selectors.html#working-with-relative-xpaths 

See consolidated commit with the updates we did, plus some more, let's go through it tomorrow: https://github.com/jalavik/hepcrawl/commit/f0657c234eeb55751a01696c1729b80b9e2e7141

hepcrawl/utils.py
```diff
@@ -69,6 +71,14 @@ def collapse_initials(name):
     return name
 
 
+def split_fullname(author):
```
jalavik: I realized that this function is not ideal, it does not preserve the information in the original string via the comma. 
```python
In [5]: split_fullname("Doe, John Magic")
Out[5]: ('Doe', 'John Magic')  # correct

In [6]: split_fullname("Doe Boe, John Magic")
Out[6]: ('Doe', 'Boe John Magic') # Wrong
```

Can you add test cases to `test_utils.py` that makes sure that the following is correct:
```python
In [5]: split_fullname("Doe, John Magic")
Out[5]: ('Doe', 'John Magic')

In [6]: split_fullname("Doe Boe, John Magic")
Out[6]: ('Doe Boe', 'John Magic')
```

hepcrawl/utils.py
```diff
@@ -69,6 +71,14 @@ def collapse_initials(name):
...
+    """Split an author name to surname and given names from the form 'foo, bar'."""
+    fullname = re.sub(r',', '', author).split()
```
jalavik: I would suggest splitting on the comma, if given, and you will have the correct split. If comma is not given, you can indeed then just take the first.

hepcrawl/spiders/alpha_spider.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def start_requests(self):
+        """You can also run the spider on local test files"""
+        if self.source_file:
+            yield Request(self.source_file)
+        elif self.start_urls:
+            for url in self.start_urls:
+                yield Request(url)
+
+    def split_fullname(self, author):
```
jalavik: You can use the `split_fullname` from `utils.py` instead of defining it here.

hepcrawl/spiders/alpha_spider.py
```diff
@@ -0,0 +1,175 @@
...
+        """If we want to split the author name to surname and given names.
+        """
+        import re
+        fullname = author.split()
+        # CAVEAT: Here we are *assuming* surname comes last
+        surname = fullname[-1]
+        given_names = " ".join(fullname[:-1])
+        return surname, given_names
+
+    def has_numbers(self, s):
```
jalavik: This function can be added to `utils.py` since it does not use the spider instance `self` for anything and thus is not tied to it.

hepcrawl/spiders/alpha_spider.py
```diff
@@ -0,0 +1,175 @@
...
+        for thesis in node.xpath(self.itertag):
+            record = HEPLoader(
+                item=HEPRecord(), selector=thesis, response=response)
+
+            # Author, affiliation, year:
+            author_line = thesis.xpath(
+                "./div[@class = 'content clearfix']//div[@class='field-item even']"
+                "/p[contains(text(),'Thesis')]/text()"
+            ).extract()
+            author_data = self.parse_author_data(author_line[0])
```
jalavik: For simplicity, I believe you can just do the author parsing in one function, since `parse_author_data` already produces a list of author structures.

jalavik: Ah, I see now that the author_data contains more things. Hmm. Perhaps `parse_author_data` can return multiple arguments then. E.g. ```return authors, thesis_type, year```

hepcrawl/spiders/alpha_spider.py
```diff
@@ -0,0 +1,175 @@
...
+            authors = self.get_authors(author_data)
+            record.add_value('authors', authors)
+            record.add_value('date_published', author_data[0]['year'])
+
+            # Abstract:
+            title, urls = self.get_title(thesis)
+            record.add_value('title', title)
+            record.add_value('urls', urls)
+            abs_paragraphs = thesis.xpath(
```
jalavik: To add more "value" to the `self.get_abstract` function (and align it with other similar functions like `get_title`), I would pass the `thesis` object to the function and do the xpath inside it.

hepcrawl/spiders/base_spider.py
```diff
@@ -0,0 +1,229 @@
...
+        record = HEPLoader(item=HEPRecord(), selector=node, response=response)
+        record.add_value('files', response.meta.get("direct_link"))
+        record.add_value('urls', response.meta.get("urls"))
+        record.add_xpath('abstract', './/dc:description/text()')
+        record.add_xpath('title', './/dc:title/text()')
+        record.add_xpath('date_published', './/dc:date/text()')
+        record.add_xpath('source', './/base_dc:collname/text()')
+        # Should this be able to scrape all kinds of publications?
+        # Now does only theses:
+        record.add_value('thesis', {'degree_type': 'PhD'})
```
jalavik: Actually, will we always harvest thesis from base? If not, this information should only be added conditionally. @annetteholtkamp

hepcrawl/spiders/base_spider.py
```diff
@@ -0,0 +1,229 @@
...
+            request.meta["urls"] = urls_in_record
+            request.meta["node"] = node
+            return request
+        else:
+            response.meta["direct_link"] = direct_link
+            response.meta["urls"] = urls_in_record
+            response.meta["node"] = node
+            return self.build_item(response)
+
+    def build_item(self, response):
```
jalavik: Ideally we will add the designated collection tags on INSPIRE in the `collections` field. See e.g. https://github.com/inspirehep/hepcrawl/blob/master/hepcrawl/spiders/wsp_spider.py#L174-L185

If you are unsure about which collections we have, see with @annetteholtkamp  (Usually we mark BOOK, THESIS, CORE, Published, ConferencePaper etc.)

hepcrawl/utils.py
```diff
@@ -85,3 +95,21 @@ def get_temporary_file(prefix="tmp_",
...
+    if not url:
+        return ""
+    resp = requests.head(url)
+    content_type = resp.headers.get("Content-Type")
+    if content_type is None:
+        raise Exception("No content-type found in URL {0}".format(url))
+    return content_type
+
+
+def parse_domain(url):
```
jalavik: Can you add a test case under `test_utils.py` for this function?

hepcrawl/utils.py
```diff
@@ -85,3 +95,21 @@ def get_temporary_file(prefix="tmp_",
             pass
         raise e
     return filepath
+
+
+def get_mime_type(url):
```
jalavik: To test this function, we need to mock requests. If you wanna give it a try, take a look how to use the `responses` library in tests: https://github.com/getsentry/responses#response-body-as-string

Then you would call `get_mime_type` with the URL you have mocked and you test to make sure the function is doing the correct things. You can then test several things, such as:

* When `Content-Type` is given (this you can mock as part of responses.add), assert that the function returns it
* When `Content-Type` is not given, assert that an exception is raised. (https://pytest.org/latest/assert.html#assertions-about-expected-exceptions)
* When `url` is an empty string ("") or None, assert that an empty string is returned.

jalavik:
When these comments are addressed I believe we are ready to merge.

hepcrawl/utils.py
```diff
@@ -69,6 +71,28 @@ def collapse_initials(name):
     return name
 
 
+def split_fullname(author, surname="first"):
```
jalavik: Thanks for fixing this and adding the test. There is just one case that will cause this code to fail. In case, author is empty string or None.

And since `surname` is sort of a boolean flag, we can change it to be a boolean. E.g. `surname_first=True` as default argument. Then people would pass False if they want surname last.

hepcrawl/spiders/alpha_spider.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def start_requests(self):
+        """You can also run the spider on local test files"""
+        if self.source_file:
+            yield Request(self.source_file)
+        elif self.start_urls:
+            for url in self.start_urls:
+                yield Request(url)
+
+    def has_numbers(self, s):
```
jalavik: This one can be added to utils, with accompanying tests.

jalavik: Like 
```python
def has_numbers(text):
      ....
```

hepcrawl/spiders/alpha_spider.py
```diff
@@ -0,0 +1,150 @@
...
+
+            abstract = self.get_abstract(thesis)
+            record.add_value("abstract", abstract)
+
+            record.add_xpath(
+                'files', "./div[@class = 'content clearfix']//span[@class='file']/a/@href")
+            record.add_value('source', 'Alpha experiment')
+            record.add_value('collections', ['HEP', 'THESIS'])
+
+            if "phd" in thesis_type.lower():
```
jalavik: To save wasted computation if the `thesis_type` is not `"phd"`, you can put the check further up and change it to use `continue`. E.g.

```python
for thesis in node.xpath(self.itertag):
    ....
    authors, thesis_type, year = self.parse_author_data(thesis)
    if "phd" not in thesis_type.lower():
        continue  # this record will be ignored
    ....
```

jalavik: Then at the bottom you would always yield.

tests/test_utils.py
```diff
@@ -108,3 +113,45 @@ def test_build_dict(list_for_dict):
...
+    """Test MIME type getting.
+
+    Apparently the mocked response always has 'Content-Type'
+    and it seems to default to 'text/html'?
+    Because of this, right now is not testing if missing header
+    information raises and exception.
+    """
+    url = 'http://www.example.com/files/einstein_1905.pdf'
+    responses.add(responses.HEAD, url, status=200, content_type='application/pdf')
+    resp = requests.head(url)
```
jalavik: Actually, I think you can remove this line and the `if` check below and just have

```python
assert get_mime_type(url) == 'application/pdf'
assert get_mime_type(None) == ''
assert get_mime_type('') == ''
```

tests/test_utils.py
```diff
@@ -108,3 +113,45 @@ def test_build_dict(list_for_dict):
...
+    domain = 'http://www.example.com/'
+    assert parse_domain(url) == domain
+
+
+@responses.activate
+def test_get_mime_type():
+    """Test MIME type getting.
+
+    Apparently the mocked response always has 'Content-Type'
+    and it seems to default to 'text/html'?
```
jalavik: Yeah, or `text/plain`, what you can do is to add another test function where in your mock you try with `content_type=None` as part of the arguments.

PR #4, opened 2015-12-01
Aps spider json

Lilykos:


hepcrawl/utils.py
```diff
@@ -85,3 +85,24 @@ def get_temporary_file(prefix="tmp_",
             pass
         raise e
     return filepath
+
+
+def get_nested(root, *keys):
```
jalavik: Can you enrich `test_utils.py` with test cases for these functions?

tests/responses/__init__.py
```diff
@@ -33,10 +33,14 @@ def fake_response_from_file(file_name, url=None):
...
 
-    response = Response(
+    response = response_type(
         url=url,
         request=request,
-        body=file_content
+        body=file_content,
     )
-    response.encoding = 'utf-8'
+    try:
```
jalavik: If you set the default `response_type` to `TextResponse` and include `encoding='utf-8` in the instantiation, we can get rid of this part. (I think we can assume that all responses we want to mock this way is going to be a `TextResponse` as `HttpResponse` and `XMLResponse` inhertits from `TextReponse`)

tests/test_aps.py
```diff
@@ -0,0 +1,172 @@
...
+        "We use a popular fictional disease, zombies, in order to introduce techniques used in modern epidemiology"
+        " modeling, and ideas and techniques used in the numerical study of critical phenomena. We consider variants of"
+        " zombie models, from fully connected continuous time dynamics to a full scale exact stochastic dynamic"
+        " simulation of a zombie outbreak on the continental United States. Along the way, we offer a closed form"
+        " analytical expression for the fully connected differential equation, and demonstrate that the single person"
+        " per site two dimensional square lattice version of zombies lies in the percolation universality class. We end"
+        " with a quantitative study of the full scale US outbreak, including the average susceptibility of different"
+        " geographical regions."
+    )
+    record = results.next()
```
jalavik: Some spider `parse` functions do not yield a generator always (but a list), and to avoid different techniques between tests, I propose to just put it in a `for` loop like this:

```python
for record in results:
    # test code
```

jalavik:
The CI run is reporting some issues:

* `check-manifest` is complaining that the JSON test file is not included, you'll need to add `recursive-include tests *.json` to `MANIFEST.in`.
* PEP8 issue:
```
/home/travis/build/inspirehep/hepcrawl/hepcrawl/utils.py:108:73: W292 no newline at end of file
```

jalavik:
Great. I just need now to ask you to merge the commits into one, and amend the commit log message to align with the rest of the commit log messages in the repo. See e.g. https://github.com/inspirehep/hepcrawl/commit/66e0661e7a85447770303c1fb9709e59eb73cf0c

Basically, amending the headline with a proper module/folder name followed by a colon and the subject:  E.g. `spiders: APS spider`

PR #11, opened 2016-02-12
loaders: output processor for DOI structure

bittirousku:
* Adds class `Dois` to outputs.py which creates structured DOIs in the JSON file.
* This is then called in `HEPLoader`. 

Signed-off-by: Henrik Vesterinen <henrik.vesterinen@cern.ch>

### DIFF
--- a/hepcrawl/loaders.py
+++ b/hepcrawl/loaders.py
@@ -29,6 +29,7 @@ from .inputs import (
 from .outputs import (
     FreeKeywords,
     ClassificationNumbers,
+    Dois,
 )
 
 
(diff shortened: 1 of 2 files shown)


jalavik:
@bittirousku Thanks! This is great. I am also doing some changes to the loaders and HEP items in my work here: https://github.com/jalavik/hepcrawl/commit/16193f598ae25255f794c22d78503f9886f35ed0 

I am sure we can still keep this as well alongside for the DOI field.

bittirousku:
@jalavik Good to know! I made a small modification: changed the source to "DOI". Is this now correct?

jalavik:
@bittirousku Actually, I think it can be the value of the `source` field, but it can be done in the push to INSPIRE, as other fields need the same. So you can omit the `source` field for now.

PR #12, opened 2016-02-17
spiders: new Elsevier spider

bittirousku:
* Includes tests.

* Should do everything the old harvestingkit is doing.


hepcrawl/loaders.py
```diff
@@ -99,3 +101,13 @@ class HEPLoader(ItemLoader):
     dois_out = ListToValueDict()
     related_article_doi_out = ListToValueDict()
     urls_out = ListToValueDict(key="url")
+
+
+class ElsevierLoader(HEPLoader):
```
jalavik: Do you think we shall have the MathML cleaning for all records perhaps? So that you just need to amend the existing `abstract_in` loader?

hepcrawl/inputs.py
```diff
@@ -75,3 +75,10 @@ def clean_tags_from_affiliations(value):
 def clean_collaborations(value):
     """Remove the prefixes for collaborations"""
     return value.replace("for the", "").strip()
+
+
+def clean_abstract(text):
```
jalavik: This can be renamed to something more generic as perhaps not only abstracts needs this kind of cleaning. Suggestion: `clean_whitespace_characters`

bittirousku:
Made the loader a bit more generic. It's a good idea to preserve MathML in other spiders also.

jalavik:
@bittirousku Thanks! From the Travis run it seems like you need to add the zip file used for testing to the branch.

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -0,0 +1,1087 @@
...
+        pub_name = ref.xpath(".//sb:publisher/sb:name/text()").extract_first()
+        pub_location = ref.xpath(
+            ".//sb:publisher/sb:location/text()").extract_first()
+        if pub_location:
+            return u"{}: {}".format(pub_location, pub_name)
+        else:
+            return pub_name
+
+    # @staticmethod
+    # def _resolve_arxiv_uri(arxiv_url):  # NOTE: do we need this?
```
jalavik: You can remove this.

bittirousku:
Added zip packages without any real content (no copyrights).

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -0,0 +1,1079 @@
...
+            book_title = ref.xpath(
+                ".//sb:book//sb:maintitle/text()").extract_first()
+        return book_title
+
+    @staticmethod
+    def _fix_node_text(text_nodes):
+        """Text might be split to multiple elements, this joins them.
+
+        Input must me a list. Returns a string."""
+        # NOTE: should this be more general?
```
jalavik: Yeah, in general you can use the loaders for the HEPLoader to join results. See e.g. `Join` http://doc.scrapy.org/en/latest/topics/loaders.html#scrapy.loader.processors.Join for the https://github.com/inspirehep/hepcrawl/blob/master/hepcrawl/loaders.py#L63-L67

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -0,0 +1,1079 @@
...
+        if textref:
+            references["raw_reference"] = self._fix_node_text(textref)
+            return references
+        doi = ref.xpath(".//ce:doi/text()").extract_first()
+        fpage = ref.xpath(".//sb:first-page/text()").extract_first()
+        lpage = ref.xpath(".//sb:last-page/text()").extract_first()
+        publication = self._get_ref_journal_title(ref)
+        title = self._get_ref_title(ref)
+        book_title = self._get_ref_book_title(ref, title)
+        # NOTE: do we need the book edition:
```
jalavik: @bittirousku @annetteholtkamp Does the book edition fit in the new reference structure?

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -0,0 +1,1079 @@
...
+        # edition = ref.xpath(".//sb:edition/text()").extract_first()
+        # TODO: shoud use Inspire standard format journal names, like this:
+        # fixed_title, _ = fix_journal_name(journal, self.journal_mappings)
+        volume = self._get_ref_volume(ref)
+        issue = ref.xpath(".//sb:issue-nr/text()").extract_first()
+        comments = ref.xpath(".//sb:comment/text()").extract()
+        comment = ", ".join([com.strip("()") for com in comments]).strip(": ")
+        isbn = ref.xpath(".//sb:isbn/text()").extract_first()
+        # NOTE do we want ISSN info:
+        # issn = ref.xpath(".//sb:issn/text()").extract_first()
```
jalavik: @annetteholtkamp Is ISSN info something we can store for references?

hepcrawl/spiders/elsevier_spider.py
bittirousku: @jalavik OK, `Join` can be used with titles. But for other cases I'm not sure this should be moved to the loader. I'm using `_fix_node_text` to process text that will not be the final product (eg. in `_get_ref_title`). Also it's removing unwanted whitespaces in addition to joining.

bittirousku:
@jalavik I moved the FFT file item structure creation to the pipeline. Is it better this way? Also when should it be decided what are the values for file 'access' and 'type'?

bittirousku:
@jalavik Tests are passing locally now, but Travis is failing.

jalavik:
@bittirousku There seem to be a subtle difference in the abstract output, `h i` instead of `hi`. It could be due to different versions of dependencies. For now. Can you check what could be different with this part of the text? And potentially just remove this "broken" part.

jalavik:
@bittirousku Excellent. Thanks!

PR #13, opened 2016-03-04
base_spider: improved title getting

bittirousku:
* Now understands subtitles correctly.


hepcrawl/spiders/base_spider.py
```diff
@@ -188,12 +200,17 @@ def build_item(self, response):
...
+        if title:
+            record.add_value('title', title)
+        if subtitle:
+            record.add_value('subtitle', subtitle)
         record.add_xpath('date_published', './/dc:date/text()')
         record.add_xpath('source', './/base_dc:collname/text()')
         record.add_value("authors", self.get_authors(node))
         record.add_value('thesis', {'degree_type': 'PhD'})
-        record.add_value('collections', ['HEP', 'THESIS'])  # When it is checked if a thesis is about HEP?
+        # NOTE: When it is checked if a thesis is about HEP:
```
jalavik: We can assume that a thesis for INSPIRE is always HEP. However, indeed from base there might be non-hep thesis. We'll have selective harvests combined with automatic classification (e.g. auto-reject) in the workflow on INSPIRE.

jalavik:
Do you have tests for this showing the handling of title and subtitles?

bittirousku:
Added test and modified the loader to take only first element in subtitle list.

jalavik:
Great thanks :+1:

PR #15, opened 2016-03-08
T2kharvest

bittirousku:
Spider to scrape T2K collaboration thesis listing (http://www.t2k.org/docs/thesis).

* Includes tests.

### DIFF
--- a/hepcrawl/spiders/t2k_spider.py
+++ b/hepcrawl/spiders/t2k_spider.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of hepcrawl.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
 # hepcrawl is a free software; you can redistribute it and/or modify it
 # under the terms of the Revised BSD License; see LICENSE file for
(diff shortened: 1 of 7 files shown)


jalavik:
@bittirousku Thanks. I will have a look at this.

hepcrawl/spiders/t2k_spider.py
```diff
@@ -0,0 +1,157 @@
...
+    http://www.t2k.org/docs/thesis
+
+    1. `parse_node` will get thesis title, author and date from the listing.
+    2. If link to the splash page exists, `scrape_for_pdf` will try to fetch
+       the pdf link and possibly also the abstract.
+    3. `build_item` will build the HEPRecord.
+
+    Example usage:
+    scrapy crawl t2k
+    scrapy crawl t2k -a source_file=file://`pwd`/tests/responses/t2k/test_list.html -s "JSON_OUTPUT_DIR=tmp/"
```
jalavik: `test_list.html` does not exist. So using e.g. `test_1.html` instead reveals the following problem:

```
2016-04-13 18:09:19 [scrapy] DEBUG: Crawled (404) <GET http://www.t2k.org/docs/thesis/001.html> (referer: file:///Users/jlavik/Envs/harvesting/src/harvest/tests/responses/t2k/test_1.html)
```
Do you see this also?

bittirousku: Aah ok, I see. The url http://www.t2k.org/docs/thesis/001 is correct. So the file name should be `001`, not `001.html`. Should this single file be added to MANIFEST? I modified everything else to work with the correct name.

Also, should I (re)add the `test_list.html`? It's the whole web page http://www.t2k.org/docs/thesis. It's not really needed anywhere as the tests are using the `001` file.

jalavik: I see. Its fine, we can just change the documentation to refer to `001.html`. 

Regarding the 404, could the html part be added in the code somewhere? (This was outside of tests btw, just a regular `scrapy crawl`).

tests/responses/t2k/test_1.html
```diff
@@ -0,0 +1,93 @@
...
+            </th>
+        </span>
+    </thead>
+    <tbody>
+        <span>
+            <tr>
+		<span>
+		   <td>
+                       <span>
+                           <a href="001.html">
```
jalavik: @bittirousku Here is perhaps why it is getting the 404 probably as it takes this link and adds `001.html`?

hepcrawl/spiders/t2k_spider.py
bittirousku: Okay. The problem was, that I renamed the original `001` file to `001.html` because *.html files were already in MANIFEST. I didn't realise it creates problems when scraping the test file. So, currently there is one line of code that ensures that file name extensions are not present when constructing splash page urls. Good/bad/ugly?

PR #16, opened 2016-03-10
dateutils: improvements to date getting [WIP]

bittirousku:
The improved date getter/formatter.  I have tested it here with Elsevier and BASE spiders.

* How should the code be organised? Right now everything's happening in `dateutils.py`.

* Should the date formatting happen in the spider or in the loader? At least with `elsevier_spider`
  it's more convenient to be inside the spider as I'm getting the year at the same time.

### DIFF
--- a/hepcrawl/spiders/elsevier_spider.py
+++ b/hepcrawl/spiders/elsevier_spider.py
@@ -16,6 +16,8 @@ import re
 from tempfile import mkdtemp
 
 import dateutil.parser as dparser
+
+import datetime
 import requests
 
 from scrapy import Request
@@ -29,6 +31,8 @@ from ..utils import (
     get_first,
     has_numbers,
 )
+
+from ..dateutils import format_year
 
 
 class ElsevierSpider(XMLFeedSpider):
@@ -326,6 +330,7 @@ class ElsevierSpider(XMLFeedSpider):
 
     def get_date(self, node):
         """Get the year, month, and day."""
+        # NOTE: this uses dateutils.py
         year = 0
         date_published = u''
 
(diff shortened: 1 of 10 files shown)


jalavik:
I think in general the date checking and parsing can happen in the loader as it is then applied equally everywhere.

If you need the year in a spider I would just parse it individually in the spider I think.

bittirousku:
Ok, as the year is extracted from the formatted date, the formatting function would be called twice. Once in the spider, once in the loader. Is that ok?

jalavik:
sure

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -797,18 +812,20 @@ def parse_node(self, response, node):
...
-            return request
+            try:
+                request = Request(sd_url, callback=self.scrape_sciencedirect)
+                request.meta["info"] = info
+                request.meta["keys_missing"] = keys_missing
+                request.meta["node"] = node
+                request.meta["xml_url"] = xml_file
+                request.meta["handle_httpstatus_list"] = self.ERROR_CODES
+                return request
+            except ValueError:
```
jalavik: Can you add a comment here explaining the reason for catching this potential exception?

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -548,24 +544,37 @@ def _get_ref_volume(ref):
         return ", ".join(volumes)
 
     @staticmethod
-    def _get_ref_years(ref):
-        """Get the reference year.
+    def range_as_string(data):
```
jalavik: Should perhaps this function be moved to utils and get a couple of tests?

jalavik:
Looks good. Thanks for this! Will merge when the minor comments are addressed. Just ping me.

bittirousku:
@jalavik I made some changes!

jalavik:
:+1:

PR #19, opened 2016-04-14
WIP  hepcrawl: new spider: PoS

ksachs:
* new spider for PoS harvesting, including sample_record and test
* modified split_fullname in utils.py
* modified affected spiders accordingly
* added language translation to utils and mappings

Signed-off-by: Kirsten Sachs <kirsten.sachs@desy.de>

PR #21, opened 2016-04-19
Proposal

ksachs:
I wanted only the branch 'proposal' - maybe I made a git mistake ...

Just to show you what I would like.
Move as much functionality to the central code (input.py / loaders.py)

The spiders get only raw information, the data-model should allow this. 
Parsing which is identical over all spiders should be done later.

E.g. for licenses any of license, license_url, or raw_license can be filled by spider, 
don't manipulate the string or derive information.
The rest incl. license_type can be filled later. 
To do this the scrapy fields have to be defined differently.

Are there tests for the functions in input.py?


jalavik:
@ksachs Thanks for this! It is a good proposal. Some tests need an update after these changes though. See full details in the test run: https://travis-ci.org/inspirehep/hepcrawl/builds/124180881

You can run tests locally by going to the root source folder and run `python setup.py test`.

Regarding git, the PoS commit is there because you added the proposal code on top of it. We could close the previous PR #19 and keep this pull request to have your PoS spider and this new code together, if thats okay?

Indeed later we can centralize the license part as well into inputs. These kinds of things will become evident probably for other fields as well as we evolve.

Actually we do not have tests for input.py (only kinda indirectly through the spider tests). It would be useful to add a test file for utils in any case.

ksachs:
in the commit massage:
* tests fail now due to wrong output values in the test functions

for phil I don't understand why the test claims the output is 
E         At index 0 diff: {'raw_name': u'Jennings, Bob'} != {u'surname': u'Jennings', u'given_names': u'Bob', u'full_name': u'Jennings, Bob'}
if I run scrapy from command-line the output is
 'authors': [{'full_name': 'Jennings, Bob',
              'given_names': u'Bob',
              'raw_name': u'Jennings, Bob',
              'surname': u'Jennings'}],

same problem for dnb

The author tests for aps, arXiv, and world_scientific I didn't change yet, since I wasn't sure that this is the output we want.

The language test in dnb I didn't change so you see that a test is useless if no one checks that the requested output value is correct.
Btw: if "en" not in languages[0].lower()
is dangerous unless you are sure it's never 'French'
https://github.com/inspirehep/hepcrawl/blob/master/hepcrawl/spiders/dnb_spider.py#L201
No cataloger can spot missing information.
 
I don't know whether I have the time to fix all tests tomorrow (esp. since I don't understand why they fail). The PoS commit is not the final version but in principle ready to merge.

jalavik:
@ksachs Thanks for pointing out the mistake in dnb spider. I will fix it.

The phil test fails as it is actually not checking the output after the record processing, but specifically the output of the `phil_spider.get_authors` function. See the "fixture" function `authors` in `test_phil.py`. It should now rather be removed and the existing author test ``test__thesis_authors`` should be amended to check the new structure.

In case it is not ready by the time you leave, I can add fixes on top later and merge things.

jalavik:
Superseded by #22

PR #28, opened 2016-06-10
spiders: new INFN spider

bittirousku:
* Scrapes INFN doctoral theses (http://www.infn.it/thesis/index.php)
* Adds tests.

### DIFF
--- a/hepcrawl/spiders/infn_spider.py
+++ b/hepcrawl/spiders/infn_spider.py
@@ -1,0 +1,230 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of hepcrawl.
+# Copyright (C) 2015, 2016 CERN.
+#
+# hepcrawl is a free software; you can redistribute it and/or modify it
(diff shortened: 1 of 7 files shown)


hepcrawl/spiders/infn_spider.py
```diff
@@ -0,0 +1,229 @@
...
+       requests. Year is given as a argument, default is current year.
+
+    2. parse_node() iterates through every record on the html page.
+
+    3. If no pdf links are found, request to scrape the splash page is returned.
+
+    4. In the end, a HEPRecord is built.
+
+
+    Example usage:
```
jalavik: We are now using Sphinx documentation language in docs. So for the examples like this, can you:

1. Wrap in examples in `    .. code-block:: console`:

```python
.. code-block:: console

    scrapy crawl infn
    etc..

```

2. Then add the spider to the spider documentation in docs/api.rst:

```
.. automodule:: hepcrawl.spiders.infn_spider
   :members:
```

PR #29, opened 2016-06-10
Iopharvest

bittirousku:
* New IOP harvester.
* Assumes local XML and PDF files, doesn't get anything from the internet (IOP STACKS is not working at the moment).


hepcrawl/extractors/nlm.py
```diff
@@ -0,0 +1,163 @@
...
+
+import datetime
+
+from ..utils import get_first
+
+
+class NLM(object):
+    """Special extractions for NLM formats."""
+
+    DOCTYPES = [
```
jalavik: Is this used somewhere?

bittirousku: At the moment no. But I'm wondering if we should do some mapping somewhere!

PR #32, opened 2016-06-14
spiders: fix issue with scrapy selectors in request.meta

bittirousku:
* It's no longer acceptable to attach selectors to scrapy requests.
* Adds new utils function `get_node`.
* The issue affected DNB and BASE spiders, they are now fixed.


jalavik:
Thanks! 👍

PR #36, opened 2016-06-16
Hindawiharvest

bittirousku:
* New spider to harvest Hindawi (http://www.hindawi.com/). This takes MARCXMLs.
* Should do some OAI harvesting to get something for this spider to eat (http://www.hindawi.com/oai-pmh/)
* Small fixes for DNB spider (It's also using MARCXML, but with different namespaces).
* Small fixes for IOP spider.

### DIFF
--- a/hepcrawl/spiders/hindawi_spider.py
+++ b/hepcrawl/spiders/hindawi_spider.py
@@ -1,0 +1,231 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of hepcrawl.
+# Copyright (C) 2016 CERN.
+#
+# hepcrawl is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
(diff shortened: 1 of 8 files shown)


jalavik:
@bittirousku Looks good! 👍   When you are ready, you can remove WIP in title and rebase and we can merge.

hepcrawl/spiders/hindawi_spider.py
```diff
@@ -0,0 +1,237 @@
...
+        return (
+            pdf_links,
+            xml_links,
+            splash_links,
+        )
+
+    @staticmethod
+    def get_abstract(node):
+        """Get abstract."""
+        return node.xpath(
```
jalavik: Any specific reason this is not directly a `record.xpath` instead of `add_value` since it is a one line xpath expression?

bittirousku: No at the moment (there was before). I'll change this.

PR #38, opened 2016-06-21
outputs: always add HEP collection

jalavik:
Signed-off-by: Jan Aage Lavik <jan.age.lavik@cern.ch>


kaplun:
Actually, in https://github.com/inspirehep/inspire-next/issues/1215 we are going to clean up situation. Any literature record is by definition 'HEP'. So we rather specify the JSON Schema to declare the type.

jalavik:
@kaplun Cool! So I guess then this is not needed.

PR #44, opened 2016-06-29
thesis: adapt to new schema

mihaibivol:
* Changes all spiders to adapt to inspirehep/inspire-next#1267

Signed-off-by: Mihai Bivol <mihai.bivol@cern.ch>

### DIFF
--- a/hepcrawl/spiders/alpha_spider.py
+++ b/hepcrawl/spiders/alpha_spider.py
@@ -76,7 +76,7 @@ class AlphaSpider(CrawlSpider):
         affiliation = ''
         for i in author_list:
             if "thesis" in i.lower():
-                thesis_type = i.strip()
+                thesis_type = re.sub(r"thesis|Thesis", "", i).strip()
             if "university" in i.lower():
                 affiliation = re.sub(r"[^A-Za-z\s]+", '', i).strip()
             if has_numbers(i):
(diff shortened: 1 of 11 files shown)


tests/test_alpha.py
```diff
@@ -111,7 +111,7 @@ def test_urls(results):
 
 
 def test_thesis(results):
-    thesis = [{'degree_type': 'PhD Thesis'}]
+    thesis = {'degree_type': 'PhD'}
```
jacquerie: Weird, how come it was passing before?

jacquerie:
LGTM, but I don't understand how come the tests were passing before.

tests/test_alpha.py
mihaibivol: I actually stripped `'thesis'` manually since it's not in the schema enum anyway. The list to value part is done via `thesis_out`

hepcrawl/spiders/alpha_spider.py
```diff
@@ -76,7 +76,7 @@ def parse_author_data(self, thesis):
         affiliation = ''
         for i in author_list:
             if "thesis" in i.lower():
-                thesis_type = i.strip()
+                thesis_type = re.sub(r"thesis|Thesis", "", i).strip()
```
mihaibivol: @jacquerie because of this thing

Issue #39, opened 2016-06-22
APS: also scrape references


jalavik:
The APS crawler is not scraping references as the JSON API do not seem to provide them:
http://harvest.aps.org/docs/harvest-api#example

So in this case the XML output might be a better source as it seem to have references. See fulltext xml example here: http://harvest.aps.org/docs/harvest-api#retrieve-all-open-access-articles

Since the crawler today works already well with the JSON format, we can adjust it to yield a secondary request per record to the XML endpoint (basically the fulltext with `Accept: text/xml`

PR #48, opened 2016-07-28
APS: add references and some other fixes

bittirousku:
- Adds support for references in APS spider.
- Adds modifications to related Elsevier spider and tests.
- Modifies BASE and DNB spider and their tests.
- Addresses issue #39


hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +65,312 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
-
-            # record.add_xpath('license', '//license/license-p/ext-link/text()')
-            # record.add_xpath('license_type', '//license/@license-type')
-            record.add_value('license_url', get_nested(article, 'rights', 'licenses')[0]['url'])
-
-            record.add_value('collections', ['HEP', 'Citeable', 'Published'])
-            yield record.load_item()
+            doi = article["identifiers"]["doi"]
+            link = "http://harvest.aps.org/v2/journals/articles/" + doi
+            # FIXME: sometimes requests are forbidden? Why??
```
kaplun: Too many requests per second.

hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +65,312 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
+        if arxiv_urls:
+            arxiv_id = arxiv_urls[0].split(":")[-1]
+            if arxiv_id and '.' not in arxiv_id:
+                return arxiv_id.strip("/")
+            else:
+                return u"arxiv:{}".format(arxiv_id)
+
+    @staticmethod
+    def _clean_text(text):
+        """Replace all occurences of certain words in a string."""
```
kaplun: What is this method trying to do?

BTW you could achieve more robust stripping of spaces using using regexp.

E.g. `re.sub(r"\s+", " ")`

kaplun:
@bittirousku why this PR is WIP?

hepcrawl/spiders/aps_spider.py
bittirousku: Trying to format the raw reference nicely. The original raw string is a mess because the XML has mixed content. One of my questions is related to this: do we want the raw reference to be a text or xml string? If we want xml string, this formatting can be removed.

bittirousku:
@kaplun This is WIP because I have a lot of questions inside the code and so far no answers.

hepcrawl/spiders/aps_spider.py
kaplun: Right! We have lately had a discussion about having hepcrawl finally producing only rawrefs, and have then them translated only at indexing/matching time. @mihaibivol @david-caro

hepcrawl/spiders/aps_spider.py
```diff
@@ -29,12 +31,17 @@ class APSSpider(Spider):
     Uses the APS REST API v2. See documentation here:
     http://harvest.aps.org/docs/harvest-api#endpoints
 
-    scrapy crawl APS -a 'from_date=2016-05-01' -a 'until_date=2016-05-15' -a 'set=openaccess'
+    Example usage:
+    .. code-block:: console
+
+        scrapy crawl APS -a 'from_date=2016-05-01' -a 'until_date=2016-05-15' -a 'set=openaccess'
+        scrapy crawl APS -a json_file=file://`pwd`/tests/responses/aps/aps_single_response.json -s "JSON_OUTPUT_DIR=tmp/"
+
```
david-caro: +1 for the examples, maybe document though the parameters too, you can do so with sphinx syntax and they will show nicely in the docs:
```
:param json_file: json file with the contents from blablabla
:type json_file: string
...
```

hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +80,289 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
+    @staticmethod
+    def _get_journal_and_section(publication):
+        """Take journal title data (with possible section) and try to fix it."""
+        # FIXME: this was moved to utils in EDP pull request
+        section = ''
+        journal_title = ''
+        possible_sections = ["A", "B", "C", "D", "E"]
+        try:
+            # filter after re.split, which may return empty elements:
+            split_pub = filter(None, re.split(r'(\W+)', publication))
```
david-caro: not a big issue though, afaik they will still be supported for a long time...

hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +80,289 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
+        #     ignore_text = ['in', 'pp', 'edited by']
+        #     if text.startswith('Vol'):
+        #         temp = re.sub(r'\D', '', text)
+        #         if temp:
+        #             volume += temp
+        #     elif len(text) > 1 and text not in ignore_text \
+        #             and not (text.isdigit() or text[:-1].isdigit()):
+        #         unstructured_text.append(text)
+        # if unstructured_text:
+        #     unstructured_text = " ".join(unstructured_text)
```
david-caro: imo the unstructured text is not needed, no, as you say the raw is enough, but I'm not 100% sure if there's any special usage for it later somewhere... @kaplun do you know?

hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +80,289 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
+        """Parse a APS JSON file into a HEP record."""
+        article = response.meta["article"]
+
+        record = HEPLoader(item=HEPRecord(), response=response)
+
+        record.add_value('dois', get_nested(article, 'identifiers', 'doi'))
+        record.add_value('page_nr', str(article.get('numPages', '')))
+        record.add_value('abstract', get_nested(article, 'abstract', 'value'))
+        record.add_value('title', get_nested(article, 'title', 'value'))
+        # record.add_value('subtitle', '')
```
david-caro: all this commented code should go away right?

hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +80,289 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
+        sublabel = ref.xpath("./object-id/text()").extract_first()
+        if sublabel:
+            reference['number'] = sublabel
+        elif label:
+            reference['number'] = label
+
+        ref_type = ref.xpath("./@publication-type").extract_first()
+        report_no = ref.xpath('.//pub-id[@pub-id-type="other"]/text()').extract_first()
+        if not report_no:
+            report_no = ref.xpath('.//pub-id[@pub-id-type="other"]/following-sibling::text()[1]').extract_first()
```
david-caro: you can split this line easily like:

```python
ref.xpath(
    './/something/something'
    '/something[else]'
)
```

david-caro:
sorry about the review process here, github changed it and I'm still getting to know how it works...

hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +80,289 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
+
+    @staticmethod
+    def _format_arxiv_id(arxiv_urls):
+        """Return an arxiv id with format 'arxiv:1407.0275'.
+
+        For old identifiers format is 'hep-ex/9908047'. Input string may or
+        may not have an 'http://' in it.
+        """
+        # FIXME: this is also in Elsevier harvest, move to utils?
+        # Or do we want this at all?
```
bittirousku: Done.

hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +80,289 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
-                next_url = next[0].href
+            next_link = links.links_by_attr_pairs([('rel', 'next')])
+            if next_link:
+                next_url = next_link[0].href
                 yield Request(next_url)
 
+    @staticmethod
+    def _get_journal_and_section(publication):
+        """Take journal title data (with possible section) and try to fix it."""
+        # FIXME: this was moved to utils in EDP pull request
```
bittirousku: Done.

hepcrawl/spiders/aps_spider.py
```diff
@@ -58,64 +80,289 @@ def __init__(self, url=None, from_date=None, until_date=None,
...
-            # record.add_xpath('license', '//license/license-p/ext-link/text()')
-            # record.add_xpath('license_type', '//license/@license-type')
-            record.add_value('license_url', get_nested(article, 'rights', 'licenses')[0]['url'])
-
-            record.add_value('collections', ['HEP', 'Citeable', 'Published'])
-            yield record.load_item()
+            doi = article["identifiers"]["doi"]
+            link = "http://harvest.aps.org/v2/journals/articles/" + doi
+            # TEST: local test file:
+            # link = "file:///home/henrikv/.virtualenvs/hepcrawl/src/hepcrawl/tests/responses/aps/aps_single_response_test_ref.xml"
```
bittirousku: Yep, fixed.

hepcrawl/spiders/aps_spider.py
bittirousku: I added the licence and keywords from the scraped XML.

hepcrawl/spiders/aps_spider.py
bittirousku: I removed it.

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -554,20 +537,28 @@ def _get_ref_years(self, ref):
     def _parse_references(self, ref, label):
         """Parse all the references."""
         reference = {}
-        textref = ref.xpath(".//ce:textref//text()").extract()
+
+        def add_items_to_reference(**kwargs):
```
david-caro: I was actually thinking of something a bit more generic, at the global level, so you can reuse the same function on all the places...

david-caro:
I'm still reviewing though!

tests/test_pipelines.py
```diff
@@ -9,31 +9,53 @@
...
     spider = aps_spider.APSSpider()
-    items = spider.parse(fake_response_from_file('aps/aps_single_response.json', response_type=TextResponse))
+    request = spider.parse(
+        fake_response_from_file('aps/aps_single_response.json')
+    ).next()
+    response = TextResponse(
+        url=request.url,
+        request=request,
+        body=aps_xml_record,
+        **{'encoding': 'utf-8'}
```
bittirousku: Jan did it like this, and I didn't want to break anything by changing his stuff.

tests/test_pipelines.py
```diff
@@ -9,31 +9,53 @@
...
 from __future__ import absolute_import, print_function, unicode_literals
 
+import os
+
+import pkg_resources
+
 import pytest
 
+from scrapy.http import TextResponse
+
```
bittirousku: No explanation, they are probably just inserted there without much thought on ordering or structuring.

tests/test_aps.py
```diff
@@ -64,9 +93,10 @@ def test_page_nr(results):
         assert 'page_nr' in record
         assert record['page_nr'] == page_nr
 
-
+@pytest.mark.xfail
 def test_free_keywords(results):
     """Test extracting free_keywords"""
+    # FIXME: where are the keywords?
```
bittirousku: This should probably be fixed for every spider in the same pull request/commit.

hepcrawl/spiders/elsevier_spider.py
bittirousku: Well, I'm using this only in two spiders. A lot of the items in a reference require more than simple if checks, so using a generic function doesn't work. At least I couldn't think anything better; feel free to advice me on that matter.

tests/test_utils.py
```diff
@@ -261,3 +262,19 @@ def test_get_journal_and_section_invalid():
...
+
+    assert format_arxiv_id(arxiv_text1) == "hep-ph/0703285"
+    assert format_arxiv_id(arxiv_text2) == "arxiv:1106.5937"
+
+
+
+
+
+
+
```
bittirousku: I fixed this, but I might have put it in a wrong commit...

hepcrawl/spiders/elsevier_spider.py
david-caro: well, just returning the initialized reference will allow you to run the same function on any part of the code, something like:

```python
def add_items_to_references(reference=None, **kwargs):
    reference = {} if reference is None else reference
    for field, value in kwargs.iteritems():
        if value:
           reference[field] = value

        return reference
```

that way you can reuse it with an existing reference or just call it to create one (that's not so nice actually... as the name does not reflect that behavior very well though), like this:

```python
myref = {'something': 234}
myref = add_items_to_reference(myref, item1=123, item2='something else',...)

# or directly with
myref = add_items_to_reference(
    something=234,
    item1=123,
    item2='something else'
)
```
for the other items, well, it gets a bit more complicated, if you want to still use that function you can try adding an extra parameter, something like 'check_func' where you pass a lambda for the test to run... if it gets more complicated, then it's not worth using that function for that key.

tests/test_utils.py
```diff
@@ -259,5 +261,26 @@ def test_get_journal_and_section_invalid():
     publication = ""
     journal_title, section = get_journal_and_section(publication)
 
-    assert journal_title == ''
-    assert section == ''
+    assert journal_title == ""
+    assert section == ""
+
+
```
bittirousku: I like the idea, but I didn't change it because I think it makes the code look unnecessarily complicated. Are there some performance improvements here? I can change it if you really want it :)

hepcrawl/spiders/base_spider.py
```diff
@@ -125,24 +124,24 @@ def get_urls_in_record(node):
...
-        return direct_link
+        return direct_links, splash_links
 
     @staticmethod
     def get_title(node):
         """Get the title and possible subtitle."""
-        title = ''
-        subtitle = ''
+        title = ""
+        subtitle = ""
```
bittirousku: I'll clean this up!

kaplun:
@david-caro, maybe that @spirosdelviniotis could inherit this PR and finalize it?

kaplun:
(After having finished WPS)

david-caro:
Of course, first things first though.

kaplun:
Yeah, I was doing a bit of spring cleaning and came towards this long standing PR :dancer:

jacquerie:

PR #53, opened 2016-09-01
spiders: new UNESP spider

bittirousku:
- Spider to scrape theses from http://repositorio.unesp.br/handle/11449/77166/browse.
- Adds tests.
- Adds `translated_title` to items.


jacquerie:

PR #58, opened 2016-10-03
Added schema validation to main spiders

david-caro:


david-caro:
This is not enough :sob:, it turns out that the generated jsons are modified again by a `post-processor` in the pipeline before sending to inspire...

tests/test_base.py
```diff
@@ -139,6 +139,9 @@ def test_get_urls(urls):
     assert urls == ["http://hdl.handle.net/1885/10005"]
 
 
+@pytest.mark.skip(
+    'This should be refactored, the url does not respond anymore.'
+)
 def test_find_direct_links(direct_links):
     """Test direct link recognising"""
     assert direct_links == []
```
bittirousku: I kind of fixed this in my PR #48 (actually I refactored everything in the tests)

kaplun:
> This is not enough :sob:, it turns out that the generated jsons are modified again by a post-processor in the pipeline before sending to inspire...

Yeah I was mentioning it IRL :)

PR #66, opened 2016-10-24
conf: move confs to it's own dir

david-caro:
- This also enable easy distribution of the same through bdist and
  sdist.

Signed-off-by: David Caro david@dcaro.es


jacquerie:

PR #69, opened 2016-10-27
WIP ÖBV harvest

bittirousku:
- ÖBV might have an API, so this could change radically. 
- This could be made an example Selenium spider.


jacquerie:

PR #71, opened 2016-11-01
WIP spiders: new STAR spider

bittirousku:
* Spider for the French STAR collection.
* Web interface for STAR is www.theses.fr


jacquerie:

PR #75, opened 2016-11-14
iop_spider: scrape metadata and fulltexts from STACKS

bittirousku:
* Uses mechanize for scraping.
* Improves NLM extractor.
* Improves tests.


jacquerie:

PR #83, opened 2017-01-12
pipelines: send data payload through api

david-caro:
* Now when using the inspire pipelines, it will create a temporary
  file to store all the results (independent from any other pipelines)
  and send that as data payload (json encoded) to the endpoint task
  plus the path to the 'persistent storage' uri, that might be slow to
  access or not even exist at the time the submit task runs but will
  eventually get created.

### DIFF
--- a/hepcrawl/pipelines.py
+++ b/hepcrawl/pipelines.py
@@ -13,9 +13,10 @@ Don't forget to add pipelines to the ITEM_PIPELINES setting
 See: http://doc.scrapy.org/en/latest/topics/item-pipeline.html
 """
 
-import os
 import datetime
 import json
+import os
+
 import requests
 
 from inspire_schemas.api import validate as validate_schema
@@ -92,6 +93,9 @@ class InspireAPIPushPipeline(object):
     def __init__(self):
         self.count = 0
 
+    def open_spider(self, spider):
+        self.results_data = []
+
     def process_item(self, item, spider):
         """Convert internal format to INSPIRE data model."""
         self.count += 1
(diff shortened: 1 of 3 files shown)


david-caro:
@jmartinm can you review please?

hepcrawl/pipelines.py
```diff
@@ -154,13 +165,25 @@ def process_item(self, item, spider):
...
 
         validate_schema(dict(item), 'hep')
+        spider.logger.debug('Validated item.')
+        self.tmpfile.write(json.dumps(item))
+        spider.logger.debug('Persisted item to temp file.')
         return item
 
     def _prepare_payload(self, spider):
         """Return payload for push."""
+        tempfile = self.tmpfile.name
```
david-caro: I can skip this temporary var and use the self.tmpfile.name directly in the open call.

hepcrawl/pipelines.py
```diff
@@ -154,13 +165,25 @@ def process_item(self, item, spider):
...
+        self.tmpfile.write(']')
+        self.tmpfile.close()
+        with open(tempfile) as results_fd:
+            results_data = results_fd.read()
+
+        shutil.rmtree(self.tmpdir)
+        results_data = json.loads(results_data)
```
david-caro: I'll swap this two.

david-caro: Better move the cleaning of the tmpdir to the _cleanup :)

jmartinm:
LGTM.

One question, is it needed to use a tmp file to write the output of the pipeline and at the end read it again to send it to the callback?

Could the result just be kept in an attribute of the class `InspireAPIPushPipeline` as, `self.results_data = []` and at the end of the pipeline json encode it without passing through file system?

david-caro:
@jmartinm Yep, that's even faster, as we have to load the whole results anyhow having the file is not really improving anything here. On it.

PR #87, opened 2017-02-16
Literature builder refactor

rikirenz:
With this PR we are adding the` literature_builder` to hepcrawl in order to create a record that is schema compatible.

### DIFF
--- a/hepcrawl/crawler2hep.py
+++ b/hepcrawl/crawler2hep.py
@@ -1,0 +1,168 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of hepcrawl.
+# Copyright (C) 2015, 2016, 2017 CERN.
+#
+# hepcrawl is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Convert a crawler record to a valid HEP record.
+
+Don't forget to add pipelines to the ITEM_PIPELINES setting
+See: http://doc.scrapy.org/en/latest/topics/item-pipeline.html
+"""
+
+from inspire_schemas.api import LiteratureBuilder
+
+
+def crawler2hep(crawler_record):
+
+    def _filter_affiliation(affiliations):
+        return [
+            affilation.get('value')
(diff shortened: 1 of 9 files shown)


hepcrawl/crawler2hep.py
```diff
@@ -0,0 +1,178 @@
...
+        elif collection == 'bookchapter':
+            builder.add_document_type('book chapter')
+        elif collection == 'conferencepaper':
+            builder.add_document_type('conference paper')
+        elif collection in document_types:
+            builder.add_document_type(collection)
+
+    if builder.obj.get('document_type') is None:
+        builder.add_document_type('article')
+
```
rikirenz: @kaplun send me this lines of code to mange the collections. Anyway I do not like a lot this way to mange the `collections` field because we can manage this also in the spiders without writing these 70 lines of code with a lot of BL inside. In fact the first idea of @kaplun was refactoring this directly in the spiders.
Anyway I am not able to do this refactor in this sprint. Should I open a issue in order to be sure that this thing is going to be refactor in the future? @david-caro

rikirenz:
depends inspirehep/inspire-schemas#81
test are failing for because:
- the builder is not inside `inspire-schemas`
- Most of the code in Hepcrawl is generating records in according to the previous versions of the schema, the code that I have added is not able to handle all the cases. For now we handle only arXiv and POS

hepcrawl/crawler2hep.py
david-caro: :+1:

david-caro:
You can xfail all the spiders that don't use the builder for now. The schemas thing should go in first yes.

michamos:
I would like to review this to make sure everything goes into the right place, but I cannot assign myself for review on this repo. So please wait for my review before merging.

david-caro:
We are xfailing everything as they are not (and were not really) ready for production yet. Fixing them now will require some work, that will most probably be removed once we migrate the rest of the spiders to the builder structure.

Current focus is arxiv + pos, once those two are ready and working we unblock the deployment of invenio3, and we can start working on the others.

PR #93, opened 2017-03-20
pipeline: return the valid hep record

david-caro:

### DIFF
--- a/hepcrawl/crawler2hep.py
+++ b/hepcrawl/crawler2hep.py
@@ -41,11 +41,11 @@ def crawler2hep(crawler_record):
 
     for abstract in crawler_record.get('abstracts', []):
         builder.add_abstract(
-            abstract=abstract.get('abstract'),
+            abstract=abstract.get('value'),
             source=abstract.get('source')
         )
 
-    for arxiv_eprint in crawler_record.get('arxiv_eprint', []):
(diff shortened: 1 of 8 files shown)


hepcrawl/crawler2hep.py
```diff
@@ -98,9 +98,9 @@ def _filter_affiliation(affiliations):
 
     try:
         builder.add_number_of_pages(
-            number_of_pages=int(crawler_record.get('page_nr'))
+            number_of_pages=int(crawler_record.get('page_nr', [])[0])
```
jacquerie: This looks scary, if `crawler_record` has no `page_nr` this blows up.

david-caro: yep, but it's catch right after (IndexError)

jacquerie: Whoops, didn't notice that.

setup.py
```diff
@@ -15,7 +15,7 @@
 
 install_requires = [
     'autosemver~=0.2',
-    'inspire-schemas~=30.0',
+    'inspire-schemas~=30.0,>=30.2.2',
```
kaplun: Should you maybe put an upper bound?

david-caro: the ~ is an upper bound for the major version.

kaplun: Sorry, I haven't noticed it was still there the `~=30.0`

PR #94, opened 2017-03-21
global: correct the preprint_date format

david-caro:


david-caro:
this is actually not ready...

PR #95, opened 2017-03-28
WSP: introduce public_notes to spider

### LINKED ISSUE #117: Introduce `public_notes` field to the WSP spider
We have to introduce missing field `public_notes` to the WSP spider and update unit tests of WSP.

spirosdelviniotis:
* Adds `public_notes` field for spiders.
* Updates WSP unit tests for `public_notes` field.
* Fixes minor indentation for WSP unit test.

Closes #117 

### DIFF
--- a/hepcrawl/extractors/jats.py
+++ b/hepcrawl/extractors/jats.py
@@ -68,6 +68,16 @@ class Jats(object):
                     free_keywords.append(keyword)
         return free_keywords, classification_numbers
 
+    def _get_public_notes(self, node):
+        """Return list of public notes from node."""
+        public_notes = []
+        for note in node.xpath('//notes[@notes-type]'):
+            for keyword in note.xpath('p/text()').extract():
+                public_notes.append({
+                    'value': keyword,
+                })
+        return public_notes
+
     def _get_authors(self, node):
         authors = []
         for contrib in node.xpath(".//contrib[@contrib-type='author']"):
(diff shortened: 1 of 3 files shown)


hepcrawl/extractors/jats.py
```diff
@@ -68,6 +68,16 @@ def _get_keywords(self, node):
...
+    def _get_public_notes(self, node):
+        """Return list of public notes from node."""
+        public_notes = []
+        for note in node.xpath('//notes[@notes-type]'):
+            for keyword in note.xpath('p/text()').extract():
+                public_notes.append({
+                    'value': keyword,
+                })
+        return public_notes
+
```
david-caro: This Jats module is used also by the edp spitder, can you add some tests on the wsp side at least, so when we go back here when `polishing` the edp spider, we don't break the wsp one?

Hmmmpf, I'll add a task (high prio) to add tests for wsp at the same level as arxiv, using the pipeline instead of just the spider to make sure that anything outputted by the spider ends up being schema compatible.

Because, on `_get_public_notes` for example, the source is missing in the public note:
    https://github.com/inspirehep/inspire-schemas/blob/d45edc53af7dd996d281e9b245b6eff32b4d19ed/inspire_schemas/records/elements/sourced_value.yml

david-caro: Can you start working on that right away if you are blocked? it should not really affect the other prs (not much) but it will give us a lot of mind calm to be checking that throughout them.

spirosdelviniotis: With https://github.com/inspirehep/hepcrawl/pull/112 and https://github.com/inspirehep/hepcrawl/pull/115 we have the extra tests at the same level as arxiv so we can continue on this PR.

david-caro:
This can be rebased now, let me know when ready :+1:

michamos:
We don't need the public notes field from WSP, as it contains 'Communicated by' information that we don't want on INSPIRE.
If a field in the publisher's metadata is named similarly to one of ours, it doesn't necessarily mean that it has the same meaning and contains the same information. It would probably save you some work if you check this kind of content issues (what fields to get, where to put them) with @annetteholtkamp and me before implementing it.

PR #96, opened 2017-04-04
global: upgrade celery to 4.0.2 version

spirosdelviniotis:


kaplun:
Why the PR? Just asking because if we merge it then we also need to merge https://github.com/inspirehep/inspire-next/pull/2138 but we are not yet fully ready there, AFAIK.

david-caro:
@salmele yep, both depend on each other, but as the celery upgrade was way more complicated than expeced, it's currently on hold.

kaplun:
Ahah! I guess you meant @kaplun :+1:

david-caro:
Hahaha, yep, sa autocompletes to salvatore xd, sorry for the noise

jacquerie:

PR #104, opened 2017-05-03
global: add testlib module for reusability

### LINKED ISSUE #100: Refactor tests code
We should move all the reusable code to inside `hepcrawl`, in a module called `testlib` or similar, specially the **celery tasks** and the **celery monitor**.

That would allow us to remove the `init.py` files and leave the tests directory as pytest likes it, to avoid name collisions if at any time we use tox:

https://docs.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules

spirosdelviniotis:
* Adds celery_monitor module to `testlib`.
* Adds tasks module to `testlib`.
* Adds custom `scrapyd` runner for coverage.
* Adds `fixtures` module instead of `__init__.py` for useful test utils.
* Re-factored `fixtures.py`.

Closes #100


tests/unit/__init__.py
```diff
@@ -1 +1,10 @@
- 
+# -*- coding: utf-8 -*-
```
david-caro: This file can be removed

spirosdelviniotis: We need that file because there are several imports inside unit tests.
E.g.  in the `tests/unit/test_alpha.py`: 
```python
from .responses import fake_response_from_file
```

david-caro: Can you move that function to the testlib module too? so we don't need the extra init file?

hepcrawl/testlib/tasks.py
```diff
@@ -11,12 +11,11 @@
 
 from __future__ import absolute_import, print_function, unicode_literals
 
-import json
-
 from six.moves.urllib.parse import urlparse
-
 from celery import Celery
 
+import json
```
david-caro: revert this change

hepcrawl/testlib/celery_monitor.py
```diff
@@ -0,0 +1,84 @@
...
+    @classmethod
+    def do_crawl(cls,
+                 app,
+                 monitor_timeout,
+                 monitor_iter_limit,
+                 crawler_instance,
+                 project='hepcrawl',
+                 spider='WSP',
+                 settings=None,
+                 **crawler_arguments):
```
david-caro: Wasn't this reformatted?

hepcrawl/testlib/celery_monitor.py
```diff
@@ -0,0 +1,84 @@
...
+
+        if settings is None:
+            settings = {}
```
david-caro: Usually, it's a bit better to use:
```python
settings = settings or {}
```

It's a bit faster (the if is not always run), and more concise, while still quite readable.

hepcrawl/testlib/celery_monitor.py
```diff
@@ -0,0 +1,84 @@
...
+
+        with cls(app, monitor_timeout=monitor_timeout, monitor_iter_limit=monitor_iter_limit) as my_monitor:
+            crawler_instance.schedule(
+                project=project,
+                spider=spider,
+                settings=settings or {},
```
david-caro: Actually, you already did set the value of `settings` before, with this line, the previous is not needed (I think it's better to leave the previous one, and here just use `settings` without `or`).

david-caro: Because having logic inline with the parameter of a call is quite harder to debug than having the logic by itself.

hepcrawl/testlib/fixtures.py
```diff
@@ -67,4 +69,4 @@ def get_node(spider, tag, response=None, text=None, rtype="xml"):
         selector = Selector(text=text, type=rtype)
     spider._register_namespaces(selector)
     node = selector.xpath(tag)
-    return node
+    return node
```
david-caro: missing newline

hepcrawl/testlib/tasks.py
```diff
@@ -14,7 +14,6 @@
 import json
 
 from six.moves.urllib.parse import urlparse
-
```
david-caro: This space was actually ok:

```
stdlib imports

3rd party imports

current package imports
```

tests/unit/test_alpha.py
```diff
@@ -10,17 +10,23 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 import pytest
+import os
```
david-caro: this is not needed right? (there's no new line using the `os` module)

tests/unit/test_alpha.py
```diff
@@ -10,17 +10,23 @@
...
 
 
 @pytest.fixture
 def results():
     """Return results generator from the Alpha spider."""
     spider = alpha_spider.AlphaSpider()
-    records = list(spider.parse(fake_response_from_file('alpha/test_1.htm')))
+    records = list(
+        spider.parse(
+            fake_response_from_file('./../../tests/unit/responses/alpha/test_1.htm')
```
david-caro: Did you forget to add the new module file on the testlib module?

Also, we should handle the path resolution on the function, so here we can just pass a simpler path, like:
```python
fake_response_from_file('alpha/test_1.htm')
```

david-caro: Just like it was before xd

tests/unit/test_dnb.py
```diff
@@ -54,19 +54,22 @@ def record(scrape_pos_page_body):
     return request.callback(response)
 
 
+@pytest.mark.skip(reason='connecting to external source')
```
david-caro: Can you add to the message in the skips a reference to the issue to fix it? (create the issue if not there yet please)

hepcrawl/testlib/celery_monitor.py
```diff
@@ -0,0 +1,87 @@
...
+        self.monitor_timeout = monitor_timeout
+        self.monitor_iter_limit = monitor_iter_limit
+
+    def __enter__(self):
+        state = self.app.events.State()
+
+        def announce_succeeded_tasks(event):
+            state.event(event)
+            task = state.tasks.get(event['uuid'])
+            print('TASK SUCCEEDED: %s[%s] %s' % (task.name, task.uuid, task.info(),))
```
david-caro: Instead of print, you should be using logging module.

hepcrawl/testlib/celery_monitor.py
```diff
@@ -0,0 +1,87 @@
...
+            tasks = self.app.AsyncResult(task.id)
+            for task in tasks.result:
+                self.results.append(task)
+            self.recv.should_stop = True
+
+        def announce_failed_tasks(event):
+            state.event(event)
+            task = state.tasks.get(event['uuid'])
+            print('TASK FAILED: %s[%s] %s' % (task.name, task.uuid, task.info(),))
```
david-caro: idem

hepcrawl/testlib/celery_monitor.py
david-caro:
```python
import logging

LOGGER = logger.get_logger(__name__)

class Bla():
    def blo():
        LOGGER.info('My message')
```

hepcrawl/testlib/fixtures.py
```diff
@@ -65,6 +64,12 @@ def get_node(spider, tag, response=None, text=None, rtype="xml"):
...
     elif text:
         selector = Selector(text=text, type=rtype)
+
     spider._register_namespaces(selector)
     node = selector.xpath(tag)
     return node
+
+
+def get_responses_dir_file(*file_names):
+    project_root_dir = os.path.dirname(os.path.abspath(__file__ + os.path.join('.', '..', '..')))
```
david-caro:
```python
def get_responses_path(*path_chunks):
"""
    :param path_chunks: Optional extra path element to suffix the responses directory with.
     Example:
         >> get_responses_path()
         '/home/myuser/hepcrawl/tests/responses'

        >> get_responses_path('one', 'two')
        '/home/myuser/hepcrawl/tests/responses/one/two'
"""
os.path.abspath(
    os.path.join(
        os.path.dirname(os.path.abspath(__file__)),
        '..',
        '..',
    )
)
```

tests/functional/WSP/test_wsp.py
```diff
@@ -19,82 +19,8 @@
...
-        with cls(app, monitor_timeout=monitor_timeout, monitor_iter_limit=monitor_iter_limit) as my_monitor:
-            crawler_instance.schedule(
-                project=project,
-                spider=spider,
-                settings=settings or {},
-                **crawler_arguments
-            )
-
-        return my_monitor.results
+from hepcrawl.testlib.tasks import app
```
david-caro: Instead of using just `app` here better use something like `as celery_app`.

hepcrawl/testlib/fixtures.py
```diff
@@ -65,6 +64,29 @@ def get_node(spider, tag, response=None, text=None, rtype="xml"):
...
+
+def get_responses_path(*path_chunks):
+    """
+    :param path_chunks: Optional extra path element to suffix the responses directory with.
+     Example:
+        >> get_responses_path()
+        '/home/myuser/hepcrawl/tests/responses'
+
+        >> get_responses_path('one', 'two')
+        '/home/myuser/hepcrawl/tests/responses/one/two'
```
david-caro: The example should be wrapped in `:` and have a newline right after (I recommend putting it last):
    http://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html

PR #112, opened 2017-05-04
tests: add wsp unit test using the full pipeline output

### LINKED ISSUE #111: Use the full pipeline output on the wsp tests
As we do on arxiv, we should be using the full pipeline output instead of just the spider, that way all the tests will be really checking that the produced data is schema compliant.

spirosdelviniotis:
* Removes duplicated validation for the item in WSP spider.
* Adds: WSP unit test that uses the full pipeline output. In that way all the tests will be really checking 
  that the produced data are schema compliant.
* Re-enables xfailed unit tests for WSP spider.

Closes #111

### DIFF
--- a/tests/unit/test_world_scientific.py
+++ b/tests/unit/test_world_scientific.py
@@ -10,7 +10,11 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 import pytest
+import os
 
+from scrapy.crawler import Crawler
+
+from hepcrawl.pipelines import InspireCeleryPushPipeline
 from hepcrawl.spiders import wsp_spider
 
 from .responses import fake_response_from_file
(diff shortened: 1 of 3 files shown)


spirosdelviniotis:
@david-caro do you think that the introduced test is well covering the #111 or should we re-factor all the existing unit tests for WSP spider?

david-caro:
Merged in https://github.com/inspirehep/hepcrawl/commit/35d4e007e46f2b23ec3060f0d9deae2a4e0e1ddc

tests/unit/test_world_scientific.py
```diff
@@ -27,6 +30,37 @@ def results():
     return records
```
spirosdelviniotis: Nice! I am going to fix it.

PR #115, opened 2017-05-05
tests: create environment handler fixture

### LINKED ISSUE #113: unit tests: create environment handler fixture
Currently the unit tests that depend on the pipeline module to generate the records, are implicitly depending on the test `tests/unit/test_pipelines.py::test_prepare_payload` to set some env variables for them before running (meaning that if you try to run them without running that one first, they will fail).

We should refactor it and add a fixture that properly sets up and cleans up the env variables (we might want mock.patch).

spirosdelviniotis:
* Adds `tests/unit/conftest.py` module for unit tests fixtures.
* Adds fixture that sets up the env variables and at the end restores the previous environment.

Closes #113

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of hepcrawl.
+# Copyright (C) 2017 CERN.
+#
+# hepcrawl is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+from __future__ import absolute_import, division, print_function
+
+import pytest
+import os
+
+
+@pytest.fixture(scope='session')
+def set_up_clean_up_env_vars():
+    env_var_backup = dict(os.environ)
+
+    os.environ['SCRAPY_JOB'] = 'scrapy_job'
+    os.environ['SCRAPY_FEED_URI'] = 'scrapy_feed_uri'
+    os.environ['SCRAPY_LOG_FILE'] = 'scrapy_log_file'
+
+    yield
+
+    os.environ = env_var_backup
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -13,14 +13,8 @@ import pytest
 import os
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture(scope='function', autouse=True)
 def set_up_clean_up_env_vars():
     env_var_backup = dict(os.environ)
-
-    os.environ['SCRAPY_JOB'] = 'scrapy_job'
-    os.environ['SCRAPY_FEED_URI'] = 'scrapy_feed_uri'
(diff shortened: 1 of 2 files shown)


tests/unit/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+import os
+
+
+@pytest.fixture(scope='session')
+def set_up_clean_up_env_vars():
+    env_var_backup = dict(os.environ)
+
+    os.environ['SCRAPY_JOB'] = 'scrapy_job'
+    os.environ['SCRAPY_FEED_URI'] = 'scrapy_feed_uri'
+    os.environ['SCRAPY_LOG_FILE'] = 'scrapy_log_file'
```
david-caro: This does not need to be here, let the test do that itself.

tests/unit/test_pipelines.py
```diff
@@ -61,13 +61,10 @@ def expected_response():
...
 
 def test_prepare_payload(
-    tmpdir, json_spider_record, spider, expected_response,
+    tmpdir, json_spider_record, spider, expected_response, set_up_clean_up_env_vars,
 ):
     """Test that the generated payload is ok."""
     _, json_record = json_spider_record
-    os.environ['SCRAPY_JOB'] = 'scrapy_job'
-    os.environ['SCRAPY_FEED_URI'] = 'scrapy_feed_uri'
-    os.environ['SCRAPY_LOG_FILE'] = 'scrapy_log_file'
```
david-caro: You should leave this here. That way, you know that this test depends on those env variables having those values.

tests/unit/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+from __future__ import absolute_import, division, print_function
+
+import pytest
+import os
+
+
+@pytest.fixture(scope='session')
```
david-caro: You should not use session as the scope, but test.
You should also put `autouse=True` so all the tests have a clean environment.

tests/unit/test_pipelines.py
```diff
@@ -61,7 +61,7 @@ def expected_response():
 
 
 def test_prepare_payload(
-    tmpdir, json_spider_record, spider, expected_response,
+    tmpdir, json_spider_record, spider, expected_response, set_up_clean_up_env_vars,
```
david-caro: As you are using the `autouse` parameter, you don't need to explicitly put in in the prameters, this is not needed, it will be used by default, everywhere ;)

tests/unit/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+import pytest
+import os
+
+
+@pytest.fixture(scope='function', autouse=True)
+def set_up_clean_up_env_vars():
+    env_var_backup = dict(os.environ)
+
+    yield
+
```
david-caro: The code is simple enough here to not need the extra spaces, not a big issue though.

Issue #128, opened 2017-05-12
Use material whenever possible


michamos:
The `arXiv` spider harvests only preprints, so there should be `material: preprint` in the output JSON whenever there is a `material` field.
For the publishers, if we know for sure whether are harvesting an article/erratum/addendum/reprint, the `material` should be set appropriately as well.

spirosdelviniotis:
@michamos @kaplun I have a question about:

> The arXiv spider harvests only preprints, so there should be material: preprint in the output JSON whenever there is a material field.

Just harvested some `arXiv` records but I can't see any `material` field or something similar.
Can you give us an example of `arXiv` record with `material` field? That would be useful.

kaplun:
@spirosdelviniotis, what @michamos meant is that it's the responsibility of the arXiv crawler to always write in the output the corresponding `material -> preprint`, because that's always the only valid value for arXiv.

spirosdelviniotis:
@kaplun Thanks!
So, will convert the `arXiv` spider to populate the `material` field with `preprint`.

kaplun:
So discussing IRL with @spirosdelviniotis looks like the `material` subfields are present in:
* `copyright`
* `dois`
* `license`
* `persistent_identifiers`
* `publication_info`

Now thinking in detail. If we have `copyright` and `license`, we should add `preprint` as `material` (because these are referring to the copy from arXiv). For `dois` and `publication_info`, we should put `publication` because for the time being the doi is provided by the user to refer to the corresponding published version. The day in which arXiv will start to mint their own DOIs they will surely expose them in a sensible way and it will be clear that that DOI will refer to the `preprint`.

For `persistent_identifier` that's currently not filled from arXiv I am guessing.

david-caro:
DOI and copyright are not being filled up either.

kaplun:
You mean arXiv does not provide this information? Are you sure?

spirosdelviniotis:
@kaplun just checked that `arXiv` provide us `doi` and not `copyright`.

kaplun:
:+1: and for future reference: here's all and only fields currently exposed by arXiv:
https://arxiv.org/OAI/arXivRaw.xsd

PR #141, opened 2017-06-06
global: support `material` field for `dois`

spirosdelviniotis:
* Adds: `HEPLoader.get_dois` method that populates `dois` field.
	         Previously field `material` inside `dois` was missing.
* Adds: adapt `HEPLoader.get_dois` to related spiders.
* Adds: support for `material` field in `dois`.
* Adds: missing `dois` testing field from `arxiv` functional tests.

Addresses #128 

### DIFF
--- a/hepcrawl/loaders.py
+++ b/hepcrawl/loaders.py
@@ -152,7 +152,6 @@ class HEPLoader(ItemLoader):
 
     classification_numbers_out = ClassificationNumbers()
 
-    dois_out = ListToValueDict()
     related_article_doi_out = ListToValueDict()
 
     thesis_supervisor_in = MapCompose(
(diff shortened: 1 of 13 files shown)


hepcrawl/loaders.py
```diff
@@ -165,5 +164,40 @@ class HEPLoader(ItemLoader):
...
+                >>> record = HEPLoader(item=HEPRecord(), selector=selector)
+                >>> record.get_dois(dois=dois_list_str)
+
+            Using `dois` and `material`::
+
+                >>> from hepcrawl.items import HEPRecord
+                >>> record = HEPLoader(item=HEPRecord(), selector=selector)
+                >>> record.get_dois(dois=dois_list_str, material='preprint')
+        """
+        if dois is None or []:
```
david-caro: You can simplify this with:
```python
if not dois:
```
Unless empty string, False and 0 are some cases you want to use (I don't think so).

hepcrawl/loaders.py
```diff
@@ -165,5 +164,40 @@ class HEPLoader(ItemLoader):
...
+            return
+
+        doi_list = []
+        for _doi in dois:
```
david-caro: would be nicer like:
```
for doi_value in dois:
    doi = {
        'value': doi_value,
    }
...
```
As then the name of the variable is more significant (and harder to confuse with `doi`).

david-caro: You could also rename the `dois` parameter there to `dois_values`, then you can use `dois` instead of `doi_list`:
```
def get_dois(self, dois_values=None, material=None):
    ...
    dois = []
    for doi_value in dois_values:
        doi = {...}
        dois.append(doi)
```

though a more pythonic way might be using an internal function:
```python
def _build_doi(doi_value, material):
    doi = {'value': doi_value}
    if material:
        doi['material'] = material

    return doi

dois = [
    _build_doi(doi_value, material)
    for doi_value in dois_values
]
```

hepcrawl/loaders.py
```diff
@@ -165,5 +164,40 @@ class HEPLoader(ItemLoader):
     )
     urls_out = ListToValueDict()
 
+    def get_dois(self, dois=None, material=None):
```
david-caro: This should not be called `get_` as it's not getting anything, that's usually reserved for functions that return an element of what you want to get (a function called `get_name` is usually expected to return a `name`, whatever that would be, to set the name, you would use `set_name` instead). And usually those functions are not expected to change the object they are called on.

It should be better named like `add_dois`, to keep the style it uses when adding other fields. If that collides with any magic it might be doing internally, `set_dois` is also a good name.

hepcrawl/spiders/arxiv_spider.py
```diff
@@ -218,3 +221,6 @@ def _get_ext_systems_number(self, node):
             'institute': 'arXiv',
             'value': node.xpath('.//identifier//text()').extract_first()
         }
+
+    def _get_dois(self, node):
```
david-caro: Here `_get_dois` makes sense, as it's returning you the dois.

PR #143, opened 2017-06-07
global: support `material` field for `license`

spirosdelviniotis:
* Re-factored `utils.get_license` method.
* Adds: support for `license.material` field.
* Re-factored `utils.get_license` to `utils.get_license_list`.
* Removes: `license_url` unused field for `license` field.
* Removes: `license_type` unused field for `license` field.
* Adds: docstring example for `license` field.
* Adds: support `license.material` field for `arxiv` functional tests.
* Adds: support `license.material` field for `arxiv` unit tests.
* Adds: support `license.material` field for `arxiv` spider.

Addresses #128

### DIFF
--- a/hepcrawl/utils.py
+++ b/hepcrawl/utils.py
@@ -261,26 +261,38 @@ def get_journal_and_section(publication):
     return journal_title, section
 
 
-def get_license(license_url='', license_text=''):
+def get_license(
+    license_url='',
+    license_text='',
+    license_material='',
+):
     """Get the license dictionary from the url or the text of the license.
 
     Args:
         license_url(str): Url of the license to generate.
         license_text(str): Text with the description of the license (sometimes is
             all we got...).
+        license_material(str): Material of the license.
 
     Returns:
         list(dict): license object to be added to the generated record, empty list
             if no license could be extracted.
     """
-    license = []
-    if license_url:
+    def _populate_license_material(license):
+        if license_material:
+            license['material'] = license_material
+        return license
+
+    def _get_license():
         license = get_license_by_url(license_url=license_url)
+        if not license:
+            license = get_license_by_text(license_text=license_text)
 
-    if not license and license_text:
-        license = get_license_by_text(license_text=license_text)
+        return license
 
-    return license
+    license = _get_license()
+
+    return [_populate_license_material(license)] if license else []
 
 
 def get_license_by_url(license_url):
@@ -296,7 +308,7 @@ def get_license_by_url(license_url):
                 license_url.strip('/'),
             )
             break
-    return [{'license': license_str, 'url': license_url}]
+    return {'license': license_str, 'url': license_url}
 
 
 def get_license_by_text(license_text):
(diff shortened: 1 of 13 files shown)


hepcrawl/utils.py
```diff
@@ -261,26 +261,42 @@ def get_journal_and_section(publication):
...
+        if not license:
+            license = get_license_by_text(license_text=license_text)
 
-    if not license and license_text:
-        license = get_license_by_text(license_text=license_text)
+        return license
 
-    return license
+    if not license_text and not license_url:
+        return []
```
david-caro: You can get rid of this if, the case where license in empty is handled correctly by `get_license_by_url` too

hepcrawl/utils.py
```diff
@@ -261,26 +261,42 @@ def get_journal_and_section(publication):
...
             all we got...).
+        license_material(str): Material of the license.
 
     Returns:
         list(dict): license object to be added to the generated record, empty list
             if no license could be extracted.
     """
-    license = []
-    if license_url:
+    def _populate_licence_material():
```
david-caro: To avoid confusion, I'd add license as a parameter here, and pass it to the function.

```python
def _populate_license_material(license):
...

license = _populate_license_material(license)
```

That way is explicit that it uses the license you have at that point, just run locally:
```shell
python -m this | grep implicit
```

hepcrawl/utils.py
spirosdelviniotis: I totally agree with this approach but, in the case that:
```python
if not license_text and not license_url:
```
is `True` then why should it run all the next code block?
(I don't think that we will have more efficiency with this but `the good practice`)

david-caro: There are 2 main reasons (in order of importance):
* Readability: Having less special cases (ifs) makes the code read better, as you don't get lost in the details, for example:

```python
def add_affiliation(author):
    if author is None:
        return

    if author == '':
        return

    affiliation = get_affiliation()
    if affiliations is None:
        return

    if affiliations == '':
        return

    if affiliations == []:
        return

    author.add_affiliation(affiliation)

# Vs
def add_affiliation(author):
    affiliation = get_affiliation()
    author.add_affiliation(affiliation)
```
    I know the example does not make much sense, but the point is that getting rid of the special cases makes it way easier to read.

* Performance: as that if is almost all the time not triggered, having to do the extra check every time is actually decreasing the performance, for example:
```python
def myfunc():
    if this_never_happens():
        return

    do_something()
    do_something_else()
```
if in that function, the if is only happening 1/10 of the times, for each 100 times it runs, it runs a total of 90*3 + 10*2 instructions, that is 290 instructions, while the same code without the if, though it does not have a shortcut:
```python
def myfunc():
    do_something()
    do_something_else()
```
runs a total of 200 instructions. It all depends on the amount of times the if is triggered of course.

spirosdelviniotis: Nice comment @david-caro ! thanks! :smile:

spirosdelviniotis:
Re-based!

PR #144, opened 2017-06-08
global: support `material` field for `copyright`

spirosdelviniotis:
* Adds: support `copyright.material` for `arxiv` functional tests.
* Adds: support `copyright.material` for `pipeline` unit tests
  (pipeline tests are using `arxiv` spider).
* Adds: support `copyright.material` for `arxiv` spider.
* Adds: minor change to comment example in `items.py`

Addresses #128

### DIFF
--- a/hepcrawl/spiders/arxiv_spider.py
+++ b/hepcrawl/spiders/arxiv_spider.py
@@ -100,6 +100,10 @@ class ArxivSpider(XMLFeedSpider):
             license_url=node.xpath('.//license//text()').extract_first()
         )
         record.add_value('license', license)
+        record.add_value(
+            'copyright_material',
+            'preprint',
+        )
 
         parsed_record = dict(record.load_item())
         return parsed_record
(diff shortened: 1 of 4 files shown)


hepcrawl/spiders/arxiv_spider.py
```diff
@@ -100,6 +100,10 @@ def parse_node(self, response, node):
             license_url=node.xpath('.//license//text()').extract_first()
         )
         record.add_value('license', license)
+        record.add_value(
+            'copyright_material',
+            'preprint',
+        )
```
david-caro: I don't think that we should be adding this to the spider, unless we add the full copyright (at least holder or statement).

@michamos wdyt?

michamos: indeed, `copyright` should only be present if there is a copyright statement. This just says: the copyrighted material we are talking about is the preprint, without saying anything about the copyright.

spirosdelviniotis: Ok! I'll fix it! :smile:

spirosdelviniotis:
As we discussed with @michamos `arXiv` [doesn't provide any information](https://arxiv.org/help/license) about `copyright`. So I am going to **Close** this PR for `arXiv`.

david-caro:
👍

PR #145, opened 2017-06-08
global: support `material` field for `publication_info`

spirosdelviniotis:
* Adds: support for `publication_info.material` field in `arxiv` unit tests.
* Adds: `pubinfo_material` field in `items.py`.
* Adds: support for `publication_info.material` field.
* Adds: support `publication_info.material` field for `arxiv` spider.
* Fixes: doc warning for `license` field.
* Adds: `make -C docs clean` command before building the docs in order
not to pass the build when retrying on travis if there are errors.

Addresses #128 

### DIFF
--- a/hepcrawl/pipelines.py
+++ b/hepcrawl/pipelines.py
@@ -93,6 +93,7 @@ class InspireAPIPushPipeline(object):
                     'page_end': item.pop('journal_lpage', ''),
                     'note': item.pop('journal_doctype', ''),
                     'pubinfo_freetext': item.pop('pubinfo_freetext', ''),
+                    'pubinfo_material': item.pop('pubinfo_material', ''),
                 }]
                 if item.get('journal_year'):
                     item['publication_info'][0]['year'] = int(
@@ -110,6 +111,7 @@ class InspireAPIPushPipeline(object):
             'journal_doctype',
             'journal_artid',
             'pubinfo_freetext',
+            'pubinfo_material',
         ])
 
         item = crawler2hep(dict(item))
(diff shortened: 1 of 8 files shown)


spirosdelviniotis:
With this PR  #145 and #144 , #143 , #141 the #128 Closes.

spirosdelviniotis:
Re-based!

tests/functional/arxiv/fixtures/arxiv_smoke_record.json
```diff
@@ -56,6 +56,11 @@
         "value": "Proc. of CRIS2015, Gallipoli, Italy, 8 pages, 17 figures, to appear\n  in Elsevier in Nuclear and Particle Physics Proceedings (2016)"
       }
     ],
+    "publication_info": [
+      {
+        "material": "publication"
+      }
+    ],
```
david-caro: We should not populate the `publication_info` if we only have `material`. We should have though an example of arxiv record with some publication info.

hepcrawl/pipelines.py
```diff
@@ -32,7 +32,8 @@ def has_publication_info(item):
         item.get('journal_fpage') or \
         item.get('journal_lpage') or \
         item.get('journal_artid') or \
-        item.get('journal_doctype')
+        item.get('journal_doctype') or \
+        item.get('pubinfo_material')
```
david-caro: The material should not be taken into account if it's only by itself, so an item that only has the material is not considered as having publication info.

spirosdelviniotis:
Ready! :smile:

PR #179, opened 2017-10-03
Upgrade schema

michamos:
## Description

This upgrades the schemas to version 51.*

## Checklist:


michamos:
Tests are still failing because too much needs to be changed on schema/dojson upgrades, and the diffs are huge so it's hard to find what has to be amended. I spent 1.5 days trying to modify the failing tests without complete success.

The problem is that they are testing too much IMHO: whether dojson produces the right output should be (and is extensively!) tested in dojson, not hepcrawl. Instead, hepcrawl should have tests asserting that dojson has been called correctly and that the output is reasonable (e.g. number of records produced is correct and they all contain, say, a `document_type` or some other required field).

@david-caro Something to discuss when you come back.

jacquerie:
> The problem is that they are testing too much IMHO

:+1:

This style of testing adds strict coupling between different parts of the system, and moreover doesn't lead to better confidence in the quality of the software.

tests/functional/desy/fixtures/ftp_server/DESY/desy_collection_records.xml
```diff
@@ -1,926 +1,363 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <collection xmlns="http://www.loc.gov/MARC21/slim">
-    <record>
-        <controlfield tag="001">111111</controlfield>
```
david-caro: Why did you get rid of the control fields?

tests/functional/desy/fixtures/desy_ftp_records_expected.json
```diff
@@ -1,2168 +1,466 @@
...
-    "inspire_categories": [
-        {
-            "source": "undefined",
-            "term": "Accelerators"
+[
+    {
+        "core": true, 
+        "documents": [
+            {
+                "url": "http://bib-pubdb1.desy.de/record/389977/files/desy-thesis-17-035.title.pdf",
```
david-caro: Is this how they will pass us the ffts? Will it always be an external url? (I was actually expecting to have a relative path in the ftp server...).

david-caro: Will this url be open to the world or will it need some authentication?

tests/functional/desy/fixtures/ftp_server/DESY/desy_collection_records.xml
michamos: I took real batchupload records. Those were inserts, so didn't have any controlfield yield.

tests/functional/desy/fixtures/desy_ftp_records_expected.json
michamos: I have no idea how they will provide files. Isn't that something you discussed with them? Here the files are publicly available theses that were provided through HTTP, but you probably want to modify the test to take files from FTP.

tests/functional/desy/fixtures/ftp_server/DESY/desy_collection_records.xml
david-caro: okok, I'll work around that 👍

tests/functional/desy/fixtures/desy_ftp_records_expected.json
david-caro: @fschwenn @ksachs Hey! can you clarify for me how will the fft files be passed to be parsed by hepcrawl? There's really no limitation, but I want to make sure that we properly support and test whatever methods you would like (passing an external url, a relative path inside the ftp, ...)

Thanks!

david-caro: I will go on with this pr using the url for now, will adapt it later.

ksachs: DESY internal stuff (PubDB) should always be public accessible URLs.

If we use the DESY spider for external stuff, as long as the 'real' spider for that publisher is not available, if depends on how the publisher provides the full-text. 
I guess we will have the real spider for those that push to our FTP server.
The rest is public URLs.

david-caro: @ksachs so, just to make sure, we will have:
* ftp relative paths (maybe for some of the publishers that push to DESY ftp)
* public urls  (for everybody else)

No?

ksachs: I think so.

david-caro: Perfect 👍

PR #180, opened 2017-10-03
Jats parser

michamos:
## Description

This continues #174 that I pushed by mistake on hepcrawl/jats-parser instead of michamos/jats-parser.

## Checklist:


michamos:
@david-caro this is ready for review. Unit tests are passing, some integration tests are failing because of the schema bumping to v51 in #179 that this PR depends on.

tests/unit/responses/world_scientific/wsp_record.xml
```diff
@@ -34,7 +34,6 @@
                 </contrib>
             </contrib-group>
             <aff id="S0219025717500060AF001"><label><sup>1</sup></label>Department, University, City, City_code 123456, <country>C. R. Country_2</country></aff>
-            <aff id="S0219025717500060EM001"><email>name_1@domain.com</email></aff>
```
michamos: it looks like this is the way WSP do emails after all. Can you change the `JatsParser.emails` property to take also this case into account?

hepcrawl/parsers/jats.py
```diff
@@ -314,10 +314,25 @@ def get_affiliation(self, id_):
...
+        """Get the emails from the node with the specified id.
+
+        Args:
+            id_(str): the value of the ``id`` attribute of the node.
+
+        Returns:
+            List[str]: the emails from the node with that id or [] if none found.
+        """
+        email_nodes = self.root.xpath('//aff[@id=$id_]/email', id_=id_)
+        return [node.root.text for node in email_nodes]
```
michamos: you shouldn't use `Selector.root` usually, as that involves dealing with the underlying `Element` node, which is wrapped by `Selector`. Instead, you can do something like
```
emails = self.root.xpath('//aff[@id=$id_]/email/text()', id_=id_).extract()
return emails
```
Be aware of https://doc.scrapy.org/en/latest/topics/selectors.html#using-text-nodes-in-a-condition though and the difference between `string(selector)` and `selector/text()` but here it should be fine.

hepcrawl/parsers/jats.py
```diff
@@ -344,10 +359,12 @@ def get_author_affiliations(self, author_node):
...
         return affiliations
 
-    @staticmethod
-    def get_author_emails(author_node):
+    def get_author_emails(self, author_node):
         """Extract an author's email addresses."""
         emails = author_node.xpath('.//email/text()').extract()
+        referred_ids = author_node.xpath('.//xref[@ref-type="aff"]/@rid').extract()
+        aff_emails = [self.get_emails_from_refs(rid) for rid in referred_ids]
+        emails = sum(aff_emails, emails)
```
michamos: `emails.extend(aff_emails)`

hepcrawl/parsers/jats.py
```diff
@@ -344,10 +359,12 @@ def get_author_affiliations(self, author_node):
 
         return affiliations
 
-    @staticmethod
-    def get_author_emails(author_node):
+    def get_author_emails(self, author_node):
         """Extract an author's email addresses."""
         emails = author_node.xpath('.//email/text()').extract()
+        referred_ids = author_node.xpath('.//xref[@ref-type="aff"]/@rid').extract()
+        aff_emails = [self.get_emails_from_refs(rid) for rid in referred_ids]
```
michamos: if you do the change suggested on the next line, you can replace this by a generator comprehension, as `list.extend` can take an iterator:
`aff_emails = (self.get_emails_from_refs(rid) for rid in referred_ids)`

hepcrawl/parsers/jats.py
```diff
@@ -314,10 +314,25 @@ def get_affiliation(self, id_):
                 no match.
         """
         affiliation_node = self.root.xpath('//aff[@id=$id_]', id_=id_)[0]
```
michamos: I would make the selector stricter here instead having the LiteratureBuilder clean up the empty affiliation later: something like `//aff[@id=$id_ and not(./email)]` should do the job.

hepcrawl/parsers/jats.py
szymonlopaciuk: It's more elegant this way, however I don't think this will work as `self.get_emails_from_refs` returns a list of strings. In theory there could be more than one email address in an affiliation field, as you said the spec is quite lax. So the question is do we need that, or we can just extract one and then this solution is fine?

hepcrawl/parsers/jats.py
```diff
@@ -363,9 +363,8 @@ def get_author_emails(self, author_node):
         """Extract an author's email addresses."""
         emails = author_node.xpath('.//email/text()').extract()
         referred_ids = author_node.xpath('.//xref[@ref-type="aff"]/@rid').extract()
-        aff_emails = [self.get_emails_from_refs(rid) for rid in referred_ids]
-        emails = sum(aff_emails, emails)
-
+        referred_emails = (self.get_emails_from_refs(rid) for rid in referred_ids)
+        emails.extend(itertools.chain(*referred_emails))
```
michamos: FYI `chain.from_iterable` is better as it doesn't build the intermediate list.

hepcrawl/parsers/jats.py
```diff
@@ -0,0 +1,529 @@
...
+    It can be used directly by invoking the :func:`parse` method, or be
+    subclassed to customize its behavior.
+
+    Args:
+        jats_record (Union[str, scrapy.selector.Selector]): the record in JATS format to parse.
+        source (Optional[str]): if provided, sets the ``source`` everywhere in
+            the record. Otherwise, the source is extracted from the JATS metadata.
+
+    Returns:
+        object: the same record in the Inspire Literature schema.
```
david-caro: This return is wrong right? I guess that it's for the `__init__` method, that will return an instance of `JatsParser`, that does not represent a record if I'm not mistaken. Probably it was moved from the parse method?

szymonlopaciuk: It does explain lack of `Returns:` in `parse` docstring, I can move it

hepcrawl/parsers/jats.py
```diff
@@ -455,3 +465,65 @@ def get_author(self, author_node):
...
+
+        for xpath, handler in fields:
+            value = citation_node.xpath(xpath).extract_first()
+            if value:
+                handler(value)
+
+        for author in self.get_reference_authors(citation_node, 'author'):
+            builder.add_author(author, 'author')
+
+        for editor in self.get_reference_authors(citation_node, 'editor'):
```
michamos: I would swap this with the authors, to have the editors before the authors in the record.

hepcrawl/utils.py
```diff
@@ -144,6 +144,28 @@ def get_first(iterable, default=None):
     return default
 
 
+def best_match(iterable, default=None):
+    """Get first truthy value from iterable, fall back to default.
```
david-caro: The name of the function looks a bit misleading, maybe it's better to call it something like `get_first` or similar, wdyt?

hepcrawl/parsers/jats.py
```diff
@@ -64,9 +65,18 @@ def parse(self):
         for keyword in self.keywords:
             self.builder.add_keyword(**keyword)
         self.builder.add_imprint_date(self.publication_date.dumps())
+        for reference in self.parse_references():
+            self.builder.add_reference(reference)
 
         return self.builder.record
 
+    def parse_references(self):
```
michamos: I would transform this into a `references` `property` like the other fields.

hepcrawl/parsers/jats.py
```diff
@@ -455,3 +465,65 @@ def get_author(self, author_node):
...
+            './person-group[@person-group-type=$role]/string-name/text()',
+            role=role
+        ).extract()
+
+
+    def get_reference(self, ref_node):
+        """Extract one reference.
+
+        Args:
+            citation_node(scrapy.selector.Selector): a selector on a single
```
michamos: wrong variable name

hepcrawl/parsers/jats.py
```diff
@@ -455,3 +465,65 @@ def get_author(self, author_node):
...
+        Returns:
+            dict: the parsed reference, as generated by
+                :class:`inspire_schemas.api.ReferenceBuilder`
+        """
+
+        builder = ReferenceBuilder()
+
+        citation_node = ref_node.xpath('./mixed-citation')
+
+        fields = [
```
michamos: you should also add the raw reference.

tests/unit/test_parsers_jats.py
```diff
@@ -159,5 +159,15 @@ def test_collaborations(aps_jats, aps_parsed):
     assert result == expected
 
 
+def test_references(aps_jats, aps_parsed):
+    result_refs = aps_jats.parse_references()
+    expected = aps_parsed['references']
+
+    assert len(result_refs) == len(expected)
+
+    for idx in range(len(expected)):
```
michamos: you can simply `assert result_refs == expected`, no need to teach `pytest` what it means for two lists to be equal.

hepcrawl/utils.py
david-caro: Somebody already did something similar xd
    https://pypi.python.org/pypi/first

(not that we sholud use this, it's simple enough here, I don't think it's worth it to add an extra dependency).

tests/unit/responses/aps/PhysRevX.7.021022_parsed.yml
```diff
@@ -24,3 +24,702 @@ license_url: https://creativecommons.org/licenses/by/4.0/
...
+      full_name: Wall, A.C.
+    title:
+      title: Reconstruction of Bulk Operators within the Entanglement Wedge in Gauge-Gravity
+        Duality
+    dois:
+    - 10.1103/PhysRevLett.117.021601
+    label: '21'
+- reference:
+    persistent_identifiers:
+    - value: arXiv:hep-th/1601.07616
```
michamos: this is wrong, an arXiv identifier is not a handle. This is probably due to a weird mixup of old and new-style identifiers. Can you amend the builder to try harder to treat this as an arXiv eprint number? the result should be `arxiv_eprint: 1601.07616`.

tests/unit/test_world_scientific.py
```diff
@@ -124,8 +124,8 @@ def test_abstract(generated_record, expected_abstract):
             [{
                 'source': 'WSP',
                 'title': (
-                    'High-efficient Solid-state Perovskite Solar Cell Without '
-                    'Lithium Salt in the Hole Transport Material'
+                    'HIGH-EFFICIENT SOLID-STATE PEROVSKITE SOLAR CELL WITHOUT '
+                    'LITHIUM SALT IN THE HOLE TRANSPORT MATERIAL'
```
david-caro: What happened here? Why the caps change?

hepcrawl/utils.py
michamos: I took the name from Flask, but :+1: to change it as it is more generic.

tests/unit/test_world_scientific.py
michamos: https://github.com/inspirehep/hepcrawl/pull/180/commits/490888a6760f03dbefcc0d7eefdbae9a098fead3 happened.

david-caro: Oh... so it was like that in the original record :/ okok

hepcrawl/parsers/jats.py
```diff
@@ -502,11 +502,11 @@ def get_reference(self, ref_node):
 
         citation_node = ref_node.xpath('./mixed-citation')
 
-        raw_ref = remove_tags(
-            citation_node[0],
-            strip='self::object-id | self::*[@specific-use="suppress-display"]'
-        ).strip()
-        builder.add_raw_reference(raw_ref, source=self.builder.source)
+        builder.add_raw_reference(
+            etree.tostring(ref_node.root).strip(),
```
michamos: you shouldn't ever have to use the `etree` API: `ref_node.extract().strip()` instead.

tests/unit/responses/aps/PhysRevX.7.021022_parsed.yml
```diff
@@ -46,9 +46,16 @@ references:
     - 10.1007/JHEP04(2015)163
   raw_refs:
   - source: American Physical Society
-    value: A. Almheiri, X. Dong, and D. Harlow, Bulk Locality and Quantum Error Correction
-      in AdS/CFT, J. High Energy Phys.04 (2015) 163.JHEPFG1029-8479
-    schema: text
+    value: "<ref id=\"c1\">\n        <label>[1]</label>\n        <mixed-citation publication-type=\"\
+      journal\"><object-id>1</object-id><person-group person-group-type=\"author\"\
+      ><string-name>A. Almheiri</string-name>, <string-name>X. Dong</string-name>,\
```
michamos: looks like you are not extracting the author names.

hepcrawl/parsers/jats.py
```diff
@@ -513,7 +513,10 @@ def get_reference(self, ref_node):
         )
 
         fields = [
-            ('./source/text()', builder.set_journal_title),
+            ('self::node()[@publication-type="journal" '
```
michamos: I think you can leave off the `self::node()`, not entirely sure though.

szymonlopaciuk: I tried and it didn't work. Neither did `.[@attr="val"]`, so this was the best I came up with.

PR #194, opened 2017-12-04
Use JATS for APS

### LINKED ISSUE #184: Use more metadata from APS
## Expected Behavior

Currently, the APS spider is only using the metadata provided by the JSON API. Instead, it should use the full data present in the bagit format, which contains the fulltext in JATS format from which more metatadata can be extracted and also figures, that we might extract for Open Access papers.

szymonlopaciuk:
## Description
For APS open access papers use the JATS format to extract more information about the papers, such as references.

## Related Issue
closes #184 and #39

## Checklist:


hepcrawl/spiders/aps_spider.py
```diff
@@ -76,53 +77,24 @@ def start_requests(self):
...
-            record.add_value('collections', ['HEP', 'Citeable', 'Published'])
-
-            parsed_item = ParsedItem(
-                record=record.load_item(),
-                record_format='hepcrawl',
-            )
-
-            yield parsed_item
+            if doi:
+                request = Request(url='{}/{}'.format(self.aps_base_url, doi),
```
michamos: you should also attach the `xml` file (as a `hidden` `fulltext`) to the record with `add_document`.

szymonlopaciuk: How do I know the `key`? Do I have to attach the file myself using `invenio-records-files`, or do we already have a facility for this somewhere in Inspire?

michamos: the key should be the file name

hepcrawl/parsers/jats.py
```diff
@@ -507,53 +516,98 @@ def get_reference(self, ref_node):
...
+        Args:
+            collaboration (string): collaboration name
 
-        return builder.obj
+        Returns:
+            string: sanitized collaboration name
+        """
+        if collaboration.startswith('(') and collaboration.endswith(')'):
+            collaboration = collaboration[1:-1]
+        collaboration = re.sub(r'\sCollaboration', '', collaboration)
```
michamos: this is already handled by the builder. The parentheses handling seems more generally useful, and should be handled there too.

PR #198, opened 2017-12-07
CDS: harvest directly from OAI-PMH

### LINKED ISSUE #197: Directly harvest oai sources
Currently in order to harvest oai services we rely on invenio-oaiharvester from an app like inspire to do it, save the result in xml files on disk, and then harvest those. It would be really nice if we can instead harvest directly from them.

## Expected Behavior
For example, for the arxiv spider, instead of harvesting from a file, we should be able to pass a server (like http://export.arxiv.org/oai2), and the parameters for the harvest (like the date from, to and the sets to take into account).

As an example of similar functionality, you can check the `inspirehep oaharvester harvest` command for parameters, we can use https://sickle.readthedocs.io/en/latest/index.html too to do the oai protocol handling.

## Current Behavior
There's no support to directly harvest from oai enabled services.

kaplun:
This depends on #200.

## Description

This implements a universal OAI-PMH spider to be used in harvesting records from OAI sources, in this instance from CDS. Drops HarvestingKit dependency from the CDS spider as inspirehep/inspire-dojson#165 added support do harvest CDS directly.

## Issues

Closes #197 and #199.

## Checklist


david-caro:
@szymonlopaciuk here's the same code we talked about yesterday, it still requires changes though, like the state management and such.

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,100 @@
...
+            url=self.url,
+            set=self.set,
+            metadata_prefix=self.metadata_prefix,
+            from_date=self.from_date,
+            until_date=self.until_date
+        ))
+        now = datetime.utcnow()
+        request = Request('oaipmh+{}'.format(self.url), self.parse)
+        yield request
+        self.state[self.alias] = self._format_date(now)
```
kaplun: @david-caro it does support state management, provided the spider is launched with the appropriate flag.

david-caro: The state is not persisted, it only exists in this instance of the spider (it should be using also the `StatefulSpider`). We have to find a way to properly persist it. The one we were using before (by enabling the job dir https://doc.scrapy.org/en/latest/topics/jobs.html#keeping-persistent-state-between-batches), does not work for parallel runs on scrapyd.

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    Implements a spider for the OAI-PMH protocol by using the Python sickle library.
+
+    In case of successful harvest (OAI-PMH crawling) the spider will remember the initial starting
+    date and will use it as `from_date` argument on the next harvest.
+    """
+    name = 'OAI-PMH'
+    state = {}
+
+    def __init__(self, url, metadata_prefix='marcxml', set=None, alias=None, from_date=None, until_date=None, granularity='YYYY-MM-DD', record_class=Record, *args, **kwargs):
```
michamos: `set` is a builtin, this should be `set_`

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def _format_date(self, datetime_object):
+        if self.granularity == 'YYYY-MM-DD':
+            return datetime_object.strftime('%Y-%m-%d')
+        elif self.granularity == 'YYYY-MM-DDThh:mm:ssZ':
+            return datetime_object.strftime('%Y-%m-%dT%H:%M:%SZ')
+        else:
+            raise RuntimeError("Invalid granularity: %s" % self.granularity)
+
+    def _make_alias(self):
```
michamos: why is this needed? you can have `tuple`s as keys of `dict`s.

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        return record.xml
+
+    def parse(self, response):
+        sickle = Sickle(self.url, class_mapping={
+            'ListRecords': self.record_class,
+            'GetRecord': self.record_class,
+        })
+        try:
+            records = sickle.ListRecords(**{
```
michamos: why not simply `key=value` instead of passing a `dict`?

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,100 @@
...
+        self.state[self.alias] = self._format_date(now)
+        logger.info("Harvesting completed. Next harvesting will resume from {}".format(self.state[self.alias]))
+
+    def parse_record(self, record):
+        """
+        This method need to be reimplemented in order to provide special parsing.
+        """
+        return record.xml
+
+    def parse(self, response):
```
michamos: `response` seems not to be used. Is it receiving no useful info from `start_requests`?

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,100 @@
...
+        for record in records:
+            yield self.parse_record(record)
+
+    def _format_date(self, datetime_object):
+        if self.granularity == 'YYYY-MM-DD':
+            return datetime_object.strftime('%Y-%m-%d')
+        elif self.granularity == 'YYYY-MM-DDThh:mm:ssZ':
+            return datetime_object.strftime('%Y-%m-%dT%H:%M:%SZ')
+        else:
+            raise RuntimeError("Invalid granularity: %s" % self.granularity)
```
michamos: I would say that this is a `ValueError`.

hepcrawl/downloaders.py
```diff
@@ -0,0 +1,22 @@
...
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Additional downloaders."""
+
+
+from scrapy.http import Response
+
+
+class DummyDownloadHandler(object):
```
michamos: this seems to be needed because for oai-pmh, you pass `oaipmh+http(s)` in the url. Cannot you just pass the real URL starting with `http(s)` instead and skip this whole download handler business?

hepcrawl/spiders/cds_spider.py
```diff
@@ -36,34 +40,29 @@ class CDSSpider(XMLFeedSpider):
...
-        marcxml_record = record_xml_output(inspire_bibrec)
-        record = create_record(marcxml_record)
-        json_record = hep.do(record)
-        parsed_item = ParsedItem(
-                record=json_record,
-                record_format='hep',
+    def parse_record(self, record):
+        response = XmlResponse(self.url, encoding='utf-8', body=record.raw)
+        selector = Selector(response, type='xml')
+        selector.remove_namespaces()
```
michamos: these three lines should probably be part of the `OAIPMHSpider`.

hepcrawl/spiders/oaipmh_spider.py
michamos: cannot you adapt the crawl once thing to work here?

hepcrawl/downloaders.py
szymonlopaciuk: The issue is that we're using Sickle to do OAI stuff, and it does all of it's downloading on it's own, only exposing an iterator on the records. Since it doesn't expose any of it's requests/responses, there is not a nice way to integrate it with Scrapy so we bypass it.

kaplun: IMHO nope, because `scrapy` would otherwise try to fetch the URL with its default methods thus not allowing me to pass the ball to `sickle`.

hepcrawl/spiders/cds_spider.py
kaplun: Not sure if in general one wants to always remove namespaces.

hepcrawl/downloaders.py
michamos: `XMLFeedSpider` seems to have no such problems...

hepcrawl/spiders/oaipmh_spider.py
kaplun: This is the dummy response that is then ignored.

hepcrawl/spiders/cds_spider.py
michamos: OK, the first two lines then :)

hepcrawl/spiders/oaipmh_spider.py
szymonlopaciuk: `from` is a restricted keyword

hepcrawl/spiders/cds_spider.py
```diff
@@ -36,34 +40,29 @@ class CDSSpider(XMLFeedSpider):
...
-    ]
-
-    def __init__(self, source_file=None, **kwargs):
-        super(CDSSpider, self).__init__(**kwargs)
-        self.source_file = source_file
 
-    def start_requests(self):
-        yield Request(self.source_file)
+    def __init__(self, from_date=None, set="forINSPIRE", *args, **kwargs):
+        super(CDSSpider, self).__init__(url='http://cds.cern.ch/oai2d', metadata_prefix='marcxml', set=set, from_date=from_date, **kwargs)
```
michamos: this is not very nice. I think it's better to have the same kind of API as [`XMLFeedSpider`](https://doc.scrapy.org/en/latest/topics/spiders.html#xmlfeedspider-example), just assigning some variables.

hepcrawl/downloaders.py
michamos: so IIUC, this whole contraption is done so that in the end, `start_requests` calls `parse`. If that's all you want, cannot you call `parse` from `start_requests` directly, given that you don't care about the response?

hepcrawl/scrapy.cfg
```diff
@@ -14,7 +14,7 @@
 default = hepcrawl.settings
 
 [deploy]
-url = http://scrapyd:6800/
+url = http://localhost:6800/
```
david-caro: why this change?

hepcrawl/spiders/oaipmh_spider.py
david-caro: crawl-once uses a local db to store key-values, maybe we would be able to use it for this, though I'm not sure if it's abusing too much.

hepcrawl/downloaders.py
szymonlopaciuk: I was trying something like this yesterday and was having issues. I didn't investigate very deeply, but I think Scrapy needs a request to begin crawling (either through `start_requests` or `url`). It expects `start_requests` to return a `Request` and doesn't allow anything.

hepcrawl/scrapy.cfg
kaplun: to run it locally (not in docker)

kaplun: It was just caught in my WIP-PR

hepcrawl/spiders/oaipmh_spider.py
david-caro: maybe should be `sets`? in plural?

hepcrawl/pipelines.py
```diff
@@ -116,8 +116,11 @@ def process_item(self, item, spider):
 
         hep_record = self._post_enhance_item(item, spider)
 
-        validate(hep_record, 'hep')
-        spider.logger.debug('Validated item by Inspire Schemas.')
+        try:
+            validate(hep_record, 'hep')
+            spider.logger.debug('Validated item by Inspire Schemas.')
+        except Exception as err:
+            spider.logger.error('ERROR in validating {}: {}'.format(hep_record, err))
```
david-caro: Make sure that this is needed, and if it is, change the `.format` by `error('message %s', message)`.

hepcrawl/spiders/cds_spider.py
```diff
@@ -9,67 +9,53 @@
...
+import logging
+from flask.app import Flask
 from dojson.contrib.marc21.utils import create_record
 from inspire_dojson.hep import hep
 
-from . import StatefulSpider
+from .oaipmh_spider import OAIPMHSpider
 from ..utils import ParsedItem
 
+logger = logging.getLogger(__name__)
```
david-caro: two spaces between imports and globals and defs/class, and globals should be all caps.

hepcrawl/spiders/cds_spider.py
```diff
@@ -9,67 +9,53 @@
...
-        json_record['$schema'] = base_uri + 'hep.json'
-        parsed_item = ParsedItem(
-                record=json_record,
-                record_format='hep',
-            )
-        return parsed_item
+        with app.app_context():
+            json_record = hep.do(record)
+            base_uri = self.settings['SCHEMA_BASE_URI']
+            json_record['$schema'] = base_uri + 'hep.json'
```
david-caro: os.path.join

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,196 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of hepcrawl.
+# Copyright (C) 2015, 2016, 2017 CERN.
```
david-caro: year here is just 2017

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,196 @@
...
+from os import path, makedirs
+
+from sickle import Sickle
+from sickle.models import Record
+from sickle.oaiexceptions import NoRecordsMatch
+
+from scrapy.http import Request, XmlResponse
+from scrapy.selector import Selector
+from . import StatefulSpider
+
```
david-caro: double spaces and uppercase globals

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,196 @@
...
+class OAIPMHSpider(StatefulSpider):
+    """
+    Implements a spider for the OAI-PMH protocol by using the Python sickle library.
+
+    In case of successful harvest (OAI-PMH crawling) the spider will remember
+    the initial starting date and will use it as `from_date` argument on the
+    next harvest.
+    """
+    name = 'OAI-PMH'
+    granularity = _Granularity.DATE
```
david-caro: Let's drop the granularity stuff, we can always implement it later if needed.

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,196 @@
...
+            logger.warning(err)
+            raise StopIteration()
+        for record in records:
+            response = XmlResponse(self.url, encoding='utf-8', body=record.raw)
+            selector = Selector(response, type='xml')
+            yield self.parse_record(selector)
+
+    def _make_alias(self):
+        return '{url}?metadataPrefix={metadata_prefix}&set={set}'.format(
+            url=self.url,
```
david-caro: don't use the url for the alias

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,196 @@
...
+                 from_date=None,
+                 until_date=None,
+                 granularity=_Granularity.DATE,
+                 record_class=Record, *args, **kwargs):
+        super(OAIPMHSpider, self).__init__(*args, **kwargs)
+        self.url = url
+        self.metadata_prefix = metadata_prefix
+        self.set = oai_set
+        self.granularity = granularity
+        self.alias = alias or self._make_alias()
```
david-caro: This property is not needed

hepcrawl/spiders/cds_spider.py
```diff
@@ -57,23 +49,13 @@ def __init__(self,
...
-            )
-            with app.app_context():
-                json_record = hep.do(record)
-                base_uri = self.settings['SCHEMA_BASE_URI']
-                json_record['$schema'] = base_uri + 'hep.json'
-            return ParsedItem(record=json_record, record_format='hep')
-        except Exception:
-            logger.exception("Error when parsing record")
-            return None
+        record = create_record(selector.xpath('.//record').extract()[0])
```
michamos: `extract()[0]` is equivalent to `extract_first()` (the former raises if the list is empty, the latter returns `None` but can be overridden).

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,196 @@
...
+        self.from_date = from_date
+        self.until_date = until_date
+        self.record_class = record_class
```
david-caro: If this record_class is not needed we should remove it.

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,196 @@
...
+        Returns:
+            Optional[dict]: last run information or None if don't exist
+        """
+        file_path = self._last_run_file_path()
+        try:
+            with open(file_path) as f:
+                last_run = json.load(f)
+                logger.info('Last run file loaded: {}'.format(repr(last_run)))
+                return last_run
+        except IOError:
```
david-caro: Here check only if the file exists (first run) but any other errors should raise

david-caro: And if it's not there, raise an exception (a subclass of `Exception` with a meaningful name), then try-catch it in whomever calls this.

tests/functional/cds/fixtures/cds.xml
```diff
@@ -0,0 +1,1480 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
david-caro: strip this examples and leave only a couple meaningful records.

tests/functional/cds/test_cds.py
```diff
@@ -32,139 +41,64 @@ def override_generated_fields(record):
...
+    config,
+):
+    crawler = get_crawler_instance(config['CRAWLER_HOST_URL'])
 
     results = CeleryMonitor.do_crawl(
         app=celery_app,
         monitor_timeout=5,
         monitor_iter_limit=100,
-        events_limit=1,
+        events_limit=2,
```
david-caro: This change might not be needed.e

tests/functional/cds/test_cds.py
```diff
@@ -32,139 +41,64 @@ def override_generated_fields(record):
...
-        app=celery_app,
-        monitor_timeout=5,
-        monitor_iter_limit=20,
-        events_limit=1,
-        crawler_instance=crawler,
-        project=set_up_local_environment.get('CRAWLER_PROJECT'),
-        spider='CDS',
-        settings={},
-        **set_up_local_environment.get('CRAWLER_ARGUMENTS')
+    gotten_results = sorted(
```
david-caro: use `hepcrawl.testlib.utils.deep_sort` instead.

tests/unit/test_oaipmh.py
```diff
@@ -0,0 +1,118 @@
...
+        run['last_run_finished_at'] = '2017-12-08T23:55:54.794969'
+    return run
+
+
+@pytest.fixture(scope='function')
+def cleanup():
+    yield
+    remove('/tmp/last_runs/OAI-PMH/2cea86bbc1d329b4273a29dc603fb8c0bb91439c.json')
+    rmdir('/tmp/last_runs/OAI-PMH')
+    rmdir('/tmp/last_runs')
```
david-caro: better use `clean_dir`

tests/unit/test_oaipmh.py
```diff
@@ -0,0 +1,118 @@
...
+def test_store_and_load_last_run(spider, cleanup):
+    now = datetime.utcnow()
+    spider._save_run(started_at=now)
+
+    file_path = spider._last_run_file_path()
+    result = override_dynamic_fields(json.load(open(file_path)))
+
+    expected = override_dynamic_fields({
+        'spider': 'OAI-PMH',
+        'url': 'http://export.arxiv.org/oai2',
```
david-caro: better put here an ip that is not external (127.0.0.1/0.0.0.0...).

tests/unit/test_oaipmh.py
```diff
@@ -0,0 +1,118 @@
...
+    yield spider
+
+
+def test_last_run_file_path(spider):
+    expected = '/tmp/last_runs/OAI-PMH/2cea86bbc1d329b4273a29dc603fb8c0bb91439c.json'
+    result = spider._last_run_file_path()
+    assert expected == result
+
+
+def test_store_and_load_last_run(spider, cleanup):
```
david-caro: name of the tests (by convention) should be `test_<thing being tested>_<test case details>`

hepcrawl/spiders/oaipmh_spider.py
```diff
@@ -0,0 +1,196 @@
...
+                    .format(self.until_date or self.granularity.format(now)))
+
+    def parse_record(self, record):
+        """
+        This method need to be reimplemented in order to provide special parsing.
+
+        Args:
+            record (scrapy.selector.Selector): selector on the parsed record
+        """
+        raise NotImplementedError()
```
david-caro: Try to use this instead/also:  https://docs.python.org/2/library/abc.html

hepcrawl/pipelines.py
szymonlopaciuk: This seemed safe to remove as I checked and scrapy does not stop unless a certain number of exceptions is raised (`CLOSESPIDER_ERRORCOUNT`) which defaults to infinity. However, I removed this and all of the records failed with `ValidationError: u'_collections' is a required property`. This seems odd as the record generation is done by DoJSON and the code looks basically the same as DESY spider. I will investigate, but maybe you know something about this.

david-caro: Nothing pops up right now, will have to check :/

michamos: are you using a version of doJSON that includes the CDS conversion? you didn't bump the version in setup.py, so maybe you are using an older version that does no preprocessing for CDS.

szymonlopaciuk: @michamos I bumped and reinstalled dependencies, I now have version 57.1.5, no change. Is there any special way I need to be using DoJSON in this case?

michamos: after discussing IRL, turns out the new API that includes CDS conversion was not used.

david-caro:
Rebase + move all the 'base' spiders to a `common` module (or similar)

hepcrawl/spiders/cds_spider.py
```diff
@@ -57,23 +49,13 @@ def __init__(self,
...
+        app = Flask('hepcrawl')
+        app.config.update(
+            self.settings.getdict('MARC_TO_HEP_SETTINGS', {})
+        )
+        with app.app_context():
+            json_record = hep.do(record)
```
michamos: this shouldn't use `hep.do` but the new `marcxml2record` API. Otherwise, the CDS conversion is not triggered.

david-caro: I think I also added this in the last PR about arXiv, as it was failing the tests, but 👍 if we don't need it.

szymonlopaciuk:
Superseded by #216

PR #200, opened 2017-12-11
tests: fix failing tests after DoJSON update

szymonlopaciuk:
## Related Issue

Due to https://github.com/inspirehep/inspire-dojson/commit/030fda8a00267a831d21264e4e8c3a57b2d04807 change DESY functional tests and some unit tests began failing. This fixes them.

## Checklist:

### DIFF
--- a/tests/functional/desy/fixtures/desy_records_ftp_expected.json
+++ b/tests/functional/desy/fixtures/desy_records_ftp_expected.json
@@ -1,6 +1,7 @@
 [
     {
-        "core": true, 
+        "core": true,
+        "citeable": true,
         "documents": [
             {
                 "url": "/tmp/file_urls/full/4467212407c5a5972dc563110d1975e30b8c4450.pdf", 
@@ -221,7 +222,8 @@
         }
     }, 
     {
-        "core": true, 
+        "core": true,
+        "citeable": true,
         "documents": [
             {
                 "url": "http://bib-pubdb1.desy.de/record/389977/files/desy-thesis-17-035.title.pdf", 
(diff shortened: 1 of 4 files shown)


jacquerie:
Why do we have these tests in the first place? They look like [change detector](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) tests to me; they don't help catch errors in DoJSON (which is extensively tested!), but they slow down development of HEPCrawl.

david-caro:
@szymonlopaciuk I'm sorry for the delay, but I disagree with @jacquerie comments, can you restore the fixtures that were being used?

jacquerie:
> but I disagree with @jacquerie comments

Can you detail your argument? I made two points:
1. These tests slow down development of HEPCrawl, because they make the status of its build depend on the output of INSPIRE-DoJSON, not just its API.
2. These tests are anyway not catching bugs, because INSPIRE-DoJSON is already extensively tested, and these big fixtures were generated by running the code and piping its output in the fixture file, effectively testing the implementation against itself.

jacquerie:
We discussed this issue IRL: these tests act as a safeguard against changes in the schema, similar to how assertions like https://github.com/inspirehep/inspire-next/blob/e56c335d72570df80a6e87f1817b0a64cdd3c2c3/tests/unit/utils/test_utils_record.py#L48 in `inspire-next` help when bumping the schemas. While they are costly to maintain, they are the only way to achieve a similar result in HEPCrawl before more unit tests are written, so this is the way to go for now.

PR #204, opened 2018-01-08
jats: scrapy handles removing namespaces now

szymonlopaciuk:
Due to scrapy/parsel#99, there is no need to cleanup namespaces in JATS. Also fixes test_references in edp unit test removing the namespace declarations from raw_references.

## Checklist:

### DIFF
--- a/hepcrawl/parsers/jats.py
+++ b/hepcrawl/parsers/jats.py
@@ -20,8 +20,6 @@ from inspire_schemas.utils import split_page_artid
 from inspire_utils.date import PartialDate
 from inspire_utils.helpers import maybe_int, remove_tags
 
-from lxml import etree  # needed until a new release of parsel is made
-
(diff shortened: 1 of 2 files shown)


michamos:
\o/ they finally released a new version. Can you also remove the now unneeded https://github.com/inspirehep/hepcrawl/blob/269bc8e7b8b163599107d4c2fdd727b0baefa541/hepcrawl/parsers/jats.py#L462?

hepcrawl/parsers/jats.py
```diff
@@ -459,7 +459,6 @@ def get_root_node(jats_record):
         else:
             root = jats_record
         root.remove_namespaces()
-        etree.cleanup_namespaces(root.root)
```
michamos: please also remove the `etree` import

PR #214, opened 2018-01-23
oai: retrieve all the records at once

david-caro:
When harvesting the whole arXiv, we are getting timed out resumption
tokens.

This avoids timing out the connection resumption token.

### DIFF
--- a/hepcrawl/spiders/common/oaipmh_spider.py
+++ b/hepcrawl/spiders/common/oaipmh_spider.py
@@ -141,7 +141,11 @@ class OAIPMHSpider(LastRunStoreSpider):
         except NoRecordsMatch as err:
             LOGGER.warning(err)
             raise StopIteration()
-        for record in records:
+
+        # Avoid timing out the resumption token
+        # TODO: implemente a storage-based solution, to be able to handle large
+        #       amounts of records.
+        for record in list(records):
             response = XmlResponse(self.url, encoding='utf-8', body=record.raw)
             selector = Selector(response, type='xml')
             yield self.parse_record(selector)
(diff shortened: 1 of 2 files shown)


hepcrawl/spiders/common/oaipmh_spider.py
```diff
@@ -131,20 +131,27 @@ def parse_record(self, record):
...
-            })
+            records = sickle.ListRecords(**params)
         except NoRecordsMatch as err:
             LOGGER.warning(err)
             raise StopIteration()
 
         # Avoid timing out the resumption token
         # TODO: implemente a storage-based solution, to be able to handle large
         #       amounts of records.
+        records = list(records):
```
michamos: syntax error: unexpected `:` :smiley:

hepcrawl/spiders/common/oaipmh_spider.py
```diff
@@ -131,20 +131,27 @@ def parse_record(self, record):
...
+        LOGGER.info(
+            'Harvested %s record for params %s',
+            len(records),
+            params,
+        )
         for record in list(records):
```
michamos: you don't need the `list` here then.

hepcrawl/spiders/common/oaipmh_spider.py
david-caro: 👍

david-caro:
> Another option would be to split the logging and give the count at the end, depending on the use-case.

Currently I'm manually debugging stuff there, and this would help me, as the requests are done per-set I want to be able to know for each request how many records we are getting. But I agree that a total count would be useful too.

I was thinking on address that when we have a clear idea on how to collect metrics.

PR #217, opened 2018-01-26
oaipmh_spider: allow harvesting single records

### LINKED ISSUE #212: OAI: allow harvesting single records (GetRecord oai verb)
Currently only allow harvesting using the `ListRecords` verb, but sometimes (usually debugging or similar manual intervention) it's useful to harvest single records.

szymonlopaciuk:
**Depends on #219 and #221.**

## Description
Harvest a single record using the `GetRecord` oai verb.

## Related Issue
This closes #212.

## Checklist:

### DIFF
--- a/hepcrawl/spiders/common/oaipmh_spider.py
+++ b/hepcrawl/spiders/common/oaipmh_spider.py
@@ -54,7 +54,7 @@ class OAIPMHSpider(LastRunStoreSpider):
         url,
         format='oai_dc',
         sets=None,
-        alias=None,
+        harvest_single=None,
         from_date=None,
         until_date=None,
         **kwargs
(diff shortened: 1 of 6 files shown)


hepcrawl/spiders/common/oaipmh_spider.py
```diff
@@ -52,7 +52,7 @@ def __init__(
         url,
         format='oai_dc',
         sets=None,
-        alias=None,
+        harvest_single=None,
```
michamos: what other parameters are allowed when `harvest_single` is set? does it make sense to have both `harvest_single` and `sets`, or `from_date`, or `until_date`?

szymonlopaciuk: No, only `format` is allowed with `harvest_single`

michamos: Then it would be good to check arguments and raise an exception (`ValueError`) if used incorrectly.

hepcrawl/spiders/common/oaipmh_spider.py
```diff
@@ -47,51 +48,63 @@ class OAIPMHSpider(LastRunStoreSpider):
...
     ):
-        super(OAIPMHSpider, self).__init__(*args, **kwargs)
+        super(OAIPMHSpider, self).__init__(**self._all_kwargs)
         self.url = url
-        self.format = format
         if isinstance(sets, string_types):
-            sets = sets.split(',')
-        self.sets = sets
-        self.from_date = from_date
-        self.until_date = until_date
```
david-caro: I thin this should be kept. I don't like 'hiding' them in the `_init_kwargs` variable, it's too implicit.

hepcrawl/spiders/common/oaipmh_spider.py
```diff
@@ -70,30 +71,55 @@ def __init__(
...
+                self.format,
+                self.identifier,
+            )
+        return self.start_requests_sets(
+            self.url,
+            self.format,
+            self.sets,
+            self.from_date,
+            self.until_date,
+        )
```
david-caro: My previous comment was unfortunate... anyhow. This is ok for now, but can you think of different ways to improves this pattern? (we can chat IRL)

PR #219, opened 2018-01-30
utils: strictly check kwargs in spiders (#218)

szymonlopaciuk:
This introduced a decorator which will disallow any keyword arguments passed to the decorated method that do not begin with an underscore sign in the decorated method. This is mainly to make errors while passing arguments to spiders immediately visible. As we cannot remove kwargs from the spiders' constructors altogether (used internally by scrapyd), with this we can ensure that we are not passing unwanted data by mistake.

Additionaly this will add all of the 'public' kwargs to an `self._init_kwargs` field in the object for easier passing and all of the arguments (including non-overloaded ones) to `self._all_kwargs`. (To make passing them forward easier.)

### DIFF
--- a/hepcrawl/utils.py
+++ b/hepcrawl/utils.py
@@ -419,8 +419,12 @@ def strict_kwargs(func):
 
         if disallowed_kwargs:
             raise TypeError(
-                'Only underscored kwargs allowed in {}. '
-                'Check {} for typos.'.format(repr(func), repr(disallowed_kwargs))
+                'Only underscored kwargs or %s allowed in %s. '
+                'Check %s for typos.' % (
+                    ', '.join(spider_fields),
+                    func,
+                    ', '.join(disallowed_kwargs),
+                )
             )
 
         return func(self, *args, **kwargs)
--- a/hepcrawl/utils.py
+++ b/hepcrawl/utils.py
@@ -392,10 +392,10 @@ def strict_kwargs(func):
     (used by scrapyd), with this we can ensure that we are not passing unwanted
     data by mistake.
 
-    Additionaly this will add all of the 'public' kwargs to an `_init_kwargs`
-    field in the object for easier passing and all of the arguments (including
-    non-overloaded ones) to `_all_kwargs`. (To make passing them forward
-    easier.)
+    Additionaly this will add all of the 'public' not-None kwargs to an
+    `_init_kwargs` field in the object for easier passing and all of the
+    arguments (including non-overloaded ones) to `_all_kwargs`.
+    (To make passing them forward easier.)
 
     Args:
         func (function): a spider method
@@ -434,6 +434,7 @@ def strict_kwargs(func):
         self._init_kwargs = {
             k: v for k, v in defaults.items()
             if not k.startswith('_') and k not in spider_fields
+               and v is not None
         }
         self._all_kwargs = defaults
         return func(self, *args, **kwargs)
--- a/hepcrawl/utils.py
+++ b/hepcrawl/utils.py
@@ -410,13 +410,6 @@ def strict_kwargs(func):
 
     allowed_arguments = defined_arguments + spider_fields
 
-    if argspec.defaults:
-        defaults = dict(
-            zip(argspec.args[-len(argspec.defaults):], argspec.defaults)
-        )
-    else:
-        defaults = {}
-
     @wraps(func)
     def wrapper(self, *args, **kwargs):
         disallowed_kwargs = [
@@ -430,13 +423,6 @@ def strict_kwargs(func):
                 'Check {} for typos.'.format(func, ', '.join(disallowed_kwargs))
             )
 
-        defaults.update(kwargs)
-        self._init_kwargs = {
-            k: v for k, v in defaults.items()
-            if not k.startswith('_') and k not in spider_fields
-               and v is not None
-        }
-        self._all_kwargs = defaults
         return func(self, *args, **kwargs)
     return wrapper
 
--- a/hepcrawl/utils.py
+++ b/hepcrawl/utils.py
@@ -420,7 +420,7 @@ def strict_kwargs(func):
         if disallowed_kwargs:
             raise TypeError(
                 'Only underscored kwargs allowed in {}. '
-                'Check {} for typos.'.format(func, ', '.join(disallowed_kwargs))
+                'Check {} for typos.'.format(repr(func), repr(disallowed_kwargs))
             )
 
         return func(self, *args, **kwargs)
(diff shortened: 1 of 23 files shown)


hepcrawl/utils.py
```diff
@@ -382,6 +384,58 @@ def get_licenses(
...
+    (used by scrapyd), with this we can ensure that we are not passing unwanted
+    data by mistake.
+
+    Additionaly this will add all of the 'public' kwargs to an `_init_kwargs`
+    field in the object for easier passing and all of the arguments (including
+    non-overloaded ones) to `_all_kwargs`. (To make passing them forward
+    easier.)
+
+    Args:
+        f (function): a spider method
```
david-caro: func

hepcrawl/utils.py
```diff
@@ -382,6 +384,58 @@ def get_licenses(
...
+                'Only underscored kwargs allowed in {}. '
+                'Check {} for typos.'.format(f, ', '.join(disallowed_kwargs))
+            )
+
+        defaults.update(kwargs)
+        self._init_kwargs = {
+            k: v for k, v in defaults.items() if not k.startswith('_')
+        }
+        self._all_kwargs = defaults
+        f(self, *args, **kwargs)
```
david-caro: `return f(...)`

tests/unit/test_utils.py
```diff
@@ -77,6 +78,17 @@ def list_for_dict():
     ]
 
 
+@pytest.fixture(scope='module')
```
david-caro: you don't need a fixture for this, you can just define the class.

tests/unit/test_utils.py
```diff
@@ -261,3 +273,15 @@ def test_get_journal_and_section_invalid():
 
     assert journal_title == ''
     assert section == ''
+
+
+def test_strict_kwargs_pass(dummy_strict_kwargs_cls):
+    """Test the `strict_kwargs` decorator allowing the kwargs."""
+    dummy = dummy_strict_kwargs_cls(a=1, b=2, d=None, _x=4, settings={'DUMMY': True})
+    assert callable(dummy.__init__)
```
david-caro: Here you should test not that it's callable, but that the params were correctly passed to it, probably changing the class to save the values passed and then asserting against it.

tests/unit/test_utils.py
```diff
@@ -261,3 +273,15 @@ def test_get_journal_and_section_invalid():
...
+
+
+def test_strict_kwargs_fail(dummy_strict_kwargs_cls):
+    """Test the `strict_kwargs` decorator disallowing some kwargs."""
+    with pytest.raises(TypeError):
+        dummy_strict_kwargs_cls(a=1, b=2, e=4)
```
david-caro: you could make this more readable by using better named vars:
```python
dummy_strict_kwargs_cls(goodparam1=1, goodparam2=2, badparam=4)
```
As in the test itself you can't really tell what's the action that it causes the error without having to go to the class definition, with the renaming, `badparam` really pops out.

tests/unit/test_utils.py
```diff
@@ -77,6 +78,17 @@ def list_for_dict():
     ]
 
 
+@pytest.fixture(scope='module')
+def dummy_strict_kwargs_cls():
+    """A sample class with @strict_kwargs constructor."""
+    class Dummy(object):
+        @strict_kwargs
+        def __init__(self, a, b=10, c=20, d=30, *args, **kwargs):
+            pass
```
david-caro: as said before (but below), you might want to store the values of the params passed.

hepcrawl/utils.py
```diff
@@ -382,6 +384,49 @@ def get_licenses(
...
+    def wrapper(self, *args, **kwargs):
+        disallowed_kwargs = [
+            key for key in kwargs
+            if not key.startswith('_') and key not in allowed_arguments
+        ]
+
+        if disallowed_kwargs:
+            raise TypeError(
+                'Only underscored kwargs allowed in {}. '
+                'Check {} for typos.'.format(func, ', '.join(disallowed_kwargs))
```
david-caro: you should either:
* use `'%s' % var`
* make sure to handle unicode

As the format function will fail if one is unicode but the other is not (while the `%` will not). For example:
```
In [5]: '{}'.format(''.join([u'é']))
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-5-054f6ba4853c> in <module>()
----> 1 '{}'.format(''.join([u'é']))

UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 0: ordinal not in range(128)

In [6]: '%s' % ''.join([u'é'])
Out[6]: u'\xe9'
```

szymonlopaciuk: Wouldn't it be a bit evil for someone to use unicode-named keyword arguments here?

david-caro: Evil indeed, but completely doable:
```
In [1]: kwargs = {'é': 123}

In [2]: def lolo(**kwargs):
   ...:     print kwargs
   ...:     

In [3]: lolo(**kwargs)
{'\xc3\xa9': 123}
```

This is more a 'good practice' than this specific case, I know you usually use format for sting interpolation ;)

hepcrawl/utils.py
```diff
@@ -420,7 +420,7 @@ def wrapper(self, *args, **kwargs):
         if disallowed_kwargs:
             raise TypeError(
                 'Only underscored kwargs allowed in {}. '
-                'Check {} for typos.'.format(func, ', '.join(disallowed_kwargs))
+                'Check {} for typos.'.format(repr(func), repr(disallowed_kwargs))
```
david-caro: why not using the "%s"?

szymonlopaciuk: Huh! I thought it was being deprecated in Python 3, TIL. I'll change it then.

david-caro: I wanted an answer more than a change ;), curious why you decided to go with this instead.

michamos: FYI you can write `{!r}` in the format to use the `repr`.

david-caro: yep, also '%r' in the '%' style formatting.

david-caro: that is not a good solution in general though (using `{!r}`), if what you want is the string representation, in this case is enough as we are logging and just care about the object contents, but for other stuff it will print the string representation of the python object instead of the string **value** of it:
```
In [3]: print '{!r}'.format(''.join([u'é']))
u'\xe9'

In [4]: print '%s' % ''.join([u'é'])
é
```

szymonlopaciuk: @david-caro ...that and `repr` seemed like an easy way out under the circumstances. Otherwise I think it would've been a mess trying to use the string representation, not knowing if it's unicode or not, and having it be python 2 & 3 compatible.

PR #226, opened 2018-02-14
desy: make scraping error JSON serializable

jacquerie:
## Description:
This is a quick fix to avoid raising an exception when spawning the
Celery task that creates the workflow object, as the JSON serializer
used by Celery doesn't know how to serialize exception objects.

A more permanent fix might involve rearchitecting the HEPCrawl items
so that they include the ``errors``, ``traceback`` and perhaps other
keys, in order for inspire-crawler to consume a uniform protocol.

## Checklist:

### DIFF
--- a/hepcrawl/spiders/desy_spider.py
+++ b/hepcrawl/spiders/desy_spider.py
@@ -287,7 +287,7 @@ class DesySpider(StatefulSpider):
                 try:
                     hep_record = marcxml2record(xml_record)
                 except Exception as e:
-                    return {'xml_record': xml_record, 'error': e,
+                    return {'xml_record': xml_record, 'error': repr(e),
                             'traceback': traceback.format_tb(sys.exc_info()[2])}
                 
             return hep_record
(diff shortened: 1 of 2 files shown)


hepcrawl/spiders/desy_spider.py
```diff
@@ -287,7 +287,7 @@ def _create_json_record(xml_record):
                 try:
                     hep_record = marcxml2record(xml_record)
                 except Exception as e:
-                    return {'xml_record': xml_record, 'error': e,
+                    return {'xml_record': xml_record, 'error': str(e),
```
michamos: wouldn't it better to use `repr` instead of `str` here?
```
In [1]: e = ValueError('foo')

In [2]: print(str(e), repr(e))
foo ValueError('foo',)
```

david-caro: I agree :)

david-caro: Though this needs a big refactor anyhow to actually use the protocol thingie and show a proper message and such, so for me it can go as is now, no biggie.

PR #235, opened 2018-03-07
utils: make traceback in ParsedItem a string instead of a list

ammirate:
Signed-off-by: Antonio Cesarano <cesarano2607@gmail.com>

## Checklist:

### DIFF
--- a/hepcrawl/utils.py
+++ b/hepcrawl/utils.py
@@ -492,8 +492,8 @@ class ParsedItem(dict):
             file_name=file_name,
         )
         parsed_item.exception = exception
-        parsed_item.traceback = traceback.format_tb(sys.exc_info()[2]),
-        parsed_item.source_data = source_data,
+        parsed_item.traceback = ''.join(traceback.format_tb(sys.exc_info()[2]))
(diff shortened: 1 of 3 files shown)


hepcrawl/utils.py
```diff
@@ -492,8 +492,8 @@ def from_exception(record_format, exception, source_data, file_name):
             file_name=file_name,
         )
         parsed_item.exception = exception
-        parsed_item.traceback = traceback.format_tb(sys.exc_info()[2]),
-        parsed_item.source_data = source_data,
+        parsed_item.traceback = ''.join(traceback.format_tb(sys.exc_info()[2]))
```
szymonlopaciuk: It's been here before, but if `from_exception` is called outside of an `except` clause this will not have the traceback information. If this is intentional we could have `exception = sys,exc_info()[1]` and get rid of a parameter, making this fact more explicit?

ammirate: Good suggestion, I'll do something more: move out the `sys.exc_info` and pass the traceback explicitly to the function.

PR #237, opened 2018-04-18
parsers: add crossref api parser

vbalbp:

## Checklist:

- [ ] I have all the information that I need (if not, move to `RFC` and look for it).
- [ ] I wrote [good commit log messages](https://github.com/torvalds/subsurface-for-dirk/blob/5f15ad5a86ada3c5e574041a5f9d85235322dabb/README#L92-L119).
- [ ] My code follows the code style of this project.


hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+
+    @property
+    def material_source(self):
+        return self.record.get("source")
+
+    @property
+    def material(self):
+        title = self.title
+        subtitle = self.subtitle
+        if 'erratum' in title.lower() or 'erratum' in subtitle.lower():
```
michamos: use `startswith` if always at the beginning

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+            material = 'erratum'
+        elif 'addendum' in title.lower() or 'addendum' in subtitle.lower():
+            material = 'addendum'
+        elif "publisher's note" in title.lower() or "publisher's note" in subtitle.lower():
+            material = 'addendum'
```
michamos: `erratum`

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+    def parent_isbn(self):
+        isbn = self.record.get("ISBN")
+        if isbn:
+            return isbn[0]
+        else:
+            return isbn
+
+    @property
+    def journal_title(self):
+        j_title = self.record.get("short-container-title")
```
michamos: it's probably better to use the full title here. This will be shortened later according to Inspire's convention which might be different from the publisher's.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+    def publisher(self):
+        return self.record.get("publisher")
+
+    @property
+    def abstract(self):
+        return self.record.get("abstract")
+
+    @property
+    def imprints(self):
+        date_parts = self.record.get("issued")
```
michamos: can you link or copy part of the docs explaining what `issued` exactly means?

tests/unit/responses/crossref/2018.3804742.yml
```diff
@@ -0,0 +1,188 @@
+---
```
michamos: can you put all the crossref responses literally as we received them? so in `json` format. `expected` is fine in `yaml`.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+        affiliations = force_list(author_node.get("affiliation"))
+
+        auth_aff = []
+        for affiliation in affiliations:
+            auth_aff.append(affiliation.get('name'))
+
+        return auth_aff
+
+
+    def get_author_uuid(self, author_node):
```
michamos: `get_author_orcid`

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+            return int(self.page_end) - int(self.page_start) or 1
+
+    @staticmethod
+    def get_author_name(author_node):
+        """Extract an author's name."""
+        author_name = ', '.join([author_node.get("family"),author_node.get("given")])
+
+        return author_name
+
+    def get_author_affiliations(self, author_node):
```
michamos: are you using `self` in the body of the function? if not, it should have a `staticmethod` decorator.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+                    if date:
+                        return date
+        return None
+
+    @property
+    def page_start(self):
+        pages = self.record.get("page")
+
+        if pages:
+            return pages.split('-')[0]
```
michamos: is it impossible to have just the starting page in `page`, but no end page?

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+    def page_end(self):
+        pages = self.record.get("page")
+
+        if pages:
+            return pages.split('-')[1]
+        else:
+            return None
+
+    @property
+    def number_of_pages(self):
```
michamos: you don't have to do this, the builder takes care of it.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+
+        return author_name
+
+    def get_author_affiliations(self, author_node):
+        """Extract an author's affiliations."""
+        affiliations = force_list(author_node.get("affiliation"))
+
+        auth_aff = []
+        for affiliation in affiliations:
+            auth_aff.append(affiliation.get('name'))
```
michamos: these three lines are equivalent to 
```
auth_aff = [affiliation.get['name'] for affiliation in affiliations]
```

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+
+        return auth_aff
+
+
+    def get_author_uuid(self, author_node):
+        """Extract an author's uuid."""
+        orcid_value = author_node.get('ORCID')
+
+        if orcid_value:
```
michamos: can you change the builder so you don't have to do this test, but pass `[('ORCID', None)]` without problem and have the builder filter it out?

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+    @property
+    def number_of_pages(self):
+        first_page = self.page_start
+        last_page = self.page_end
+
+        if first_page and last_page:
+            return int(self.page_end) - int(self.page_start) or 1
+
+    @staticmethod
+    def get_author_name(author_node):
```
michamos: it should be called `author_key` instead of `author_node` everywhere, as we are dealing with a `dict`, not XML.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+        """Extract an author's name."""
+        author_name = ', '.join([author_node.get("family"),author_node.get("given")])
```
michamos: missing a space after the `,` in the `join`.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+        """
+        author_name = self.get_author_name(author_node)
+        affiliations = self.get_author_affiliations(author_node)
+        uuid = self.get_author_uuid(author_node)
+
+        return self.builder.make_author(author_name, raw_affiliations=affiliations, ids=uuid)
+
+    @property
+    def authors(self):
+        author_nodes = self.record.get("author")
```
michamos: `authors_key`, it's one `key` for several `authors`.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+
+    @property
+    def abstract(self):
+        return self.record.get("abstract")
+
+    @property
+    def imprints(self):
+        date_parts = self.record.get("issued")
+
+        if date_parts:
```
michamos: the way this is usually done to reduce the "staircase effect" is as follows:
```
if not date_parts:
    return None

# ...
```

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+            date = date_parts.get("date-parts")[0]
+            if None not in date:
+                return '-'.join(map(str,date))
```
michamos: this is not how you should do it. Have a look at `PartialDate` from `inspire-utils.date`.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+            else:
+                date_parts = self.record.get("created")
```
michamos: wasn't `created` about the DOI creation date? if so, you shouldn’t use it.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+        isbn = ref_node.get("ISBN")
+        if isbn:
+            builder.add_uid(isbn)
+        doi = ref_node.get("DOI")
+        if doi:
+            builder.add_uid(doi)
+
+        author = ref_node.get("author")
+        if author:
+            builder.add_author(author, 'author')
```
michamos: are you sure about this? I would expect this to contain several authors.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,387 @@
...
+        
+        yield builder.obj
```
michamos: you should also use the `key` as a `label`.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,351 @@
...
+        """
+        license = {
+            'imposing': self.publisher,
+            'material': self.material,
+            'url': self.get_license_url(license_key),
+        }
+
+        return license
+
+    def get_license_url(seld, license_key):
```
michamos: `self` instead of `seld`, but you don't need the `self` here, so should be decorated with `@staticmethod`.

hepcrawl/parsers/crossref.py
```diff
@@ -0,0 +1,351 @@
...
+    def get_author_affiliations(author_key):
+        """Extract an author's affiliations."""
+        affiliations = force_list(author_key.get("affiliation"))
+
+        auth_aff = [affiliation.get('name') for affiliation in affiliations]
+
+        return auth_aff
+
+
+    def get_author_orcid(self, author_key):
```
michamos: `@staticmethod`

PR #239, opened 2018-05-02
This fixes crawlonce for a known file

david-caro:
There are two main fixes:
* In case of errors, send the results to the inspire-crawler
  submit_results task even if there were no records parsed.
  That was used to be able to test the current issue.

* Functional tests: in case of timeout waiting for a task to run,
  don't trow exception, just return empty results.

* Ignore the expected exception `scrapy.exceptions.IgnoreRequest`
  when populating the errors list.

https://its.cern.ch/jira/browse/INSPIR-228

### DIFF
--- a/hepcrawl/pipelines.py
+++ b/hepcrawl/pipelines.py
@@ -113,7 +113,10 @@ class InspireAPIPushPipeline(object):
         """
         self.count += 1
         item.record = item.to_hep(source=spider.source)
-        spider.logger.debug('Got post-enhanced hep record:\n%s' % pprint.pformat(item.record))
+        spider.logger.debug(
+            'Got post-enhanced hep record:\n%s',
+            pprint.pformat(item.record),
+        )
         crawl_result = CrawlResult.from_parsed_item(item).to_dict()
         self.results_data.append(crawl_result)
         return crawl_result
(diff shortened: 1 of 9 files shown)


david-caro:
@ammirate ready for review :)

hepcrawl/pipelines.py
```diff
@@ -184,21 +192,34 @@ def close_spider(self, spider):
...
+        if 'SCRAPY_JOB' not in os.environ:
+            self._cleanup(spider)
+            return
 
         if hasattr(spider, 'tmp_dir'):
             shutil.rmtree(path=spider.tmp_dir, ignore_errors=True)
 
-        if 'SCRAPY_JOB' in os.environ and self.count > 0:
+        if hasattr(spider, 'state'):
+            got_errors = len(spider.state.get('errors', [])) > 0
```
szymonlopaciuk: Why not just `got_errors = spider.state.get('errors')`?

hepcrawl/testlib/celery_monitor.py
```diff
@@ -81,7 +89,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             timeout=self.monitor_timeout,
             wakeup=True,
         )
-        self._wait_for_results(events_iter)
+
+        try:
+            self._wait_for_results(events_iter)
+        except SocketTimeout:
+            pass
```
ammirate: Shell we log something for this particular case?

david-caro: I don't think so, as what we want is just to wait see if there are any results sent, and if not, just return empty. It's not a non-expected case.

hepcrawl/pipelines.py
david-caro: done

PR #241, opened 2018-05-08
setup: bump celery to version ~4.0

jacquerie:
## Description:
It should be noted that `hepcrawl` has been tested against version 4 of Celery since that version was released due to how the constraint is expressed, so nothing is expected to break here.

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of hepcrawl.
-# Copyright (C) 2015, 2016, 2017 CERN.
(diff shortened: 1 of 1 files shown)


david-caro:
> Though I don't understand why hepcrawl needs celery in the first place. The whole hepcrawl/inspire-crawler/workflow interaction is very confusing.


So it needs celery to be able to trigger a task, the flow is:
hepcrawl -> celery task -> inspire-crawler -> workflow

inspire-crawler is the part on inspire-next that receives the crawl results and generates workflows from them. Celery is just the protocol used between hepcrawl and inspire-crawler.

PR #242, opened 2018-05-11
global: change the config for celery 4

jacquerie:
## Description:
https://github.com/inspirehep/hepcrawl/pull/241 forgot to update the configuration for Celery 4, which led to a failure to perform the harvest on QA and https://gitlab.cern.ch/ai/it-puppet-hostgroup-inspire/commit/bb030ed9dea6a231cdedde4d8aa1349f92e91425 to fix it. ~~This PR aims to make the above commit unnecessary.~~ I wasn't able to use `CELERY_BROKER_URL` instead of `BROKER_URL`. I don't know the reason, but all this means is that the above commit is here to stay.

## Checklist:

### DIFF
--- a/hepcrawl/testlib/tasks.py
+++ b/hepcrawl/testlib/tasks.py
@@ -20,7 +20,7 @@ from celery import Celery
 
 class Config(object):
     BROKER_URL = 'pyamqp://guest:guest@rabbitmq:5672//'
-    CELERY_RESULT_BACKEND = 'amqp://guest:guest@rabbitmq:5672//'
+    CELERY_RESULT_BACKEND = 'redis://redis:6379/1'
 
 
 app = Celery()
--- a/hepcrawl/testlib/tasks.py
+++ b/hepcrawl/testlib/tasks.py
@@ -19,11 +19,8 @@ from celery import Celery
 
 
 class Config(object):
-    CELERY_RESULT_BACKEND = "amqp://guest:guest@rabbitmq:5672//"
-    BROKER_URL = "amqp://guest:guest@rabbitmq:5672//"
-    CELERY_ALWAYS_EAGER = True
-    CELERY_CACHE_BACKEND = 'memory'
-    CELERY_EAGER_PROPAGATES_EXCEPTIONS = True
+    BROKER_URL = 'pyamqp://guest:guest@rabbitmq:5672//'
+    CELERY_RESULT_BACKEND = 'amqp://guest:guest@rabbitmq:5672//'
 
 
 app = Celery()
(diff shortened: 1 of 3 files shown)


hepcrawl/pipelines.py
```diff
@@ -176,7 +176,7 @@ def __init__(self):
 
     def open_spider(self, spider):
         self.celery.conf.update(dict(
-            BROKER_URL=spider.settings['BROKER_URL'],
+            CELERY_BROKER_URL=spider.settings['CELERY_BROKER_URL'],
```
michamos: does this require reverting https://gitlab.cern.ch/ai/it-puppet-hostgroup-inspire/commit/bb030ed9dea6a231cdedde4d8aa1349f92e91425 ?

jacquerie: Eventually yes, although I will also need to fix the typo I made in https://gitlab.cern.ch/ai/it-puppet-hostgroup-inspire/commit/bb030ed9dea6a231cdedde4d8aa1349f92e91425#ece1b7b92ff70b0371eddaad468a2895e0fc8bb7_260_260.

hepcrawl/testlib/tasks.py
```diff
@@ -19,11 +19,8 @@
 
 
 class Config(object):
-    CELERY_RESULT_BACKEND = "amqp://guest:guest@rabbitmq:5672//"
-    BROKER_URL = "amqp://guest:guest@rabbitmq:5672//"
-    CELERY_ALWAYS_EAGER = True
-    CELERY_CACHE_BACKEND = 'memory'
-    CELERY_EAGER_PROPAGATES_EXCEPTIONS = True
```
jacquerie: These three configuration variables have not had effect for a long time, as they are not valid or ignored in Celery 4, so we can safely remove them.

hepcrawl/testlib/tasks.py
```diff
@@ -19,11 +19,8 @@
 
 
 class Config(object):
-    CELERY_RESULT_BACKEND = "amqp://guest:guest@rabbitmq:5672//"
-    BROKER_URL = "amqp://guest:guest@rabbitmq:5672//"
-    CELERY_ALWAYS_EAGER = True
```
michamos: I'm surprised the tests don't require some flavor of this, but they seem to pass, so it must be OK.

jacquerie: This happens to work because of the Celery Monitor, which essentially attaches receivers that listen to signals like "a task completed successfully": https://github.com/inspirehep/hepcrawl/blob/b753c2250f27b649ad5c903ff1d3c0de9229c5ac/hepcrawl/testlib/celery_monitor.py#L76-L82.

BTW, as remarked in https://github.com/inspirehep/hepcrawl/pull/241, `hepcrawl` has been tested for a long time against Celery 4, and the name of this parameter changed to `task_always_eager`, so this configuration option has had no effect for who knows how much time.

michamos:
@ammirate Don't know how this affects your e2e tests

PR #243, opened 2018-05-17
tests: fix DoJSON exceptions

drjova:
Signed-off-by: Harris Tzovanakis <me@drjova.com>


michamos:
The change of the exception type should have been a major version in hindsight.

PR #250, opened 2018-08-30
hepcrawl: arxiv spider - bug fix collaborations

ksachs:
* minimal viable product - an advanced solution would be preferable
* deals with e.g. arXiv:1808.04927

Signed-off-by: Kirsten Sachs <kirsten.sachs@desy.de>

## Description
dealing with another variation of `<author><keyname>:</keyname>`

## Motivation and Context
authors don't end up in the collaboration field

## Checklist:

- [ ] I wrote [good commit log messages](https://github.com/torvalds/subsurface-for-dirk/blob/5f15ad5a86ada3c5e574041a5f9d85235322dabb/README#L92-L119).
- [ ] My code follows the code style of this project.

### DIFF
--- a/hepcrawl/items.py
+++ b/hepcrawl/items.py
@@ -221,6 +221,9 @@ class HEPRecord(scrapy.Item):
     preprint_date = scrapy.Field()
     """Date of preprint release."""
 
+    private_notes = scrapy.Field()
+    """Private Notes. E.g. ``Warning: check collaborations``."""
+
     public_notes = scrapy.Field()
     """Notes
 
--- a/hepcrawl/items.py
+++ b/hepcrawl/items.py
@@ -222,7 +222,7 @@ class HEPRecord(scrapy.Item):
     """Date of preprint release."""
 
     private_notes = scrapy.Field()
-    """Private Notes. E.g. ``Warning: check collaborations``."""
+    """Private Notes. E.g. Warning."""
 
     public_notes = scrapy.Field()
     """Notes
--- a/hepcrawl/spiders/arxiv_spider.py
+++ b/hepcrawl/spiders/arxiv_spider.py
@@ -15,6 +15,7 @@ import re
 
 from hepcrawl.spiders.common.oaipmh_spider import OAIPMHSpider
 from scrapy import Selector
+from six.moves import zip_longest
 
 from ..items import HEPRecord
 from ..loaders import HEPLoader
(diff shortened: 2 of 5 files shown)


ksachs:
The first commit (MVP) can be merged. The second (WIP) will currently not work. Both need more complicated test records than the current ones.

ksachs:
for details see https://github.com/inspirehep/hepcrawl/issues/251

michamos:
@ksachs I rewrote the code in a way that allows me to understand what's going on (explicit look ahead instead of setting flags for the next iteration of the loop and removal of double negation), which took me the better part of the afternoon.

I think the warning part is OK if afterwards someone is looking at them. If the goal is to make decisions in the workflow based on the content of the private notes, we should figure out a better way because it will be very brittle.

For skipping the reinterpretation of previous authors as collaborations, I am not sure I understand the intent. Without it (like it was before), we very rarely get many authors in the collaboration field, which should be super obvious. With it, we very rarely get a collaboration name lost in the middle of the author list, which should be harder to find. Besides, in case there are no affiliations, the extraction of the collaboration might be wrong (part of it might be taken as an author name) so a collaboration might disappear, which is not obviously related to your warning message.

ksachs:
@michamos thanks!
Now I don't understand the code anymore. Seems to be the best way to prevent me from trying to contribute to the code :-) Anyhow, as long as all tests in test_arxiv_all.py pass (`bash -c "py.test tests/unit -vv && make -C docs clean && make -C docs html && python setup.py sdist && ls dist/*"`) it will be fine.

The main problem is that arXiv has no metadata field for collaborations, only for authornames and affiliations. Depending on what the user puts in the author field at upload time the collaboration information can end up anywhere. The code takes care of the cases I know of.

The skipping the reinterpretation of previous authors tries to take care of e.g.
http://export.arxiv.org/oai2?verb=GetRecord&identifier=oai:arXiv.org:1603.05976&metadataPrefix=arXiv
``` xml
<authors>
<author><keyname>Array</keyname><forenames>Keck</forenames></author>
<author><keyname>Collaborations</keyname><forenames>BICEP2</forenames></author>
<author><keyname>:</keyname></author>
<author><keyname>Ade</keyname><forenames>P.</forenames></author>
<author><keyname>Ahmed</keyname><forenames>Z.</forenames></author>
```
I can try to rely on `Collaborations` instead of `Collaboration` but somehow I have to know that _Array, Keck_ is not an author but a collaboration.

Since we can't predict what kind of stuff comes in if there is a `<author><keyname>:</keyname></author>` the idea is to tell the cataloger to have another look. The is no plan to do something automatically in the workflow.

Having lots of authors in the collaboration field is super obvious, and makes the metadata useless.
Better to loose a collaboration and tell the cataloger where to look for it.
Whether the warning text is meaningful for all catalogers - we will see.

PR #255, opened 2018-11-08
parsers: add arXiv parser

vbalbp:
Parser for the arXiv spider

### DIFF
--- a/hepcrawl/parsers/arxiv.py
+++ b/hepcrawl/parsers/arxiv.py
@@ -7,7 +7,7 @@
 # under the terms of the Revised BSD License; see LICENSE file for
 # more details.
 
-"""Extracts for various metadata formats"""
+"""Parser for the arXiv metadata format"""
 
 from __future__ import absolute_import, division, print_function
 
@@ -215,7 +215,7 @@ class ArxivParser(object):
     def abstract(self):
         abstract = self.root.xpath('.//abstract/text()').extract_first()
 
-        return self.fix_abstract(abstract)
+        return self.fix_long_text(abstract)
 
     @property
     def authors(self):
@@ -251,8 +251,6 @@ class ArxivParser(object):
     @property
     def number_of_pages(self):
         comments = '; '.join(self.root.xpath('.//comments/text()').extract())
-        if not comments:
-            return None
 
         found_pages = RE_PAGES.search(comments)
         if found_pages:
@@ -276,15 +274,11 @@ class ArxivParser(object):
 
     @property
     def title(self):
-        return self.fix_title(self.root.xpath('.//title/text()').extract_first())
+        return self.fix_long_text(self.root.xpath('.//title/text()').extract_first())
 
     @staticmethod
-    def fix_title(title):
-        return title.replace('\n','').replace('  ',' ')
-
-    @staticmethod
-    def fix_abstract(text):
-        return text.replace('\n',' ').replace('  ',' ')
+    def fix_long_text(text):
+        return re.sub('\s+', ' ', text).strip()
 
     @staticmethod
     def get_root_node(arxiv_record):
@@ -335,18 +329,13 @@ class ArxivParser(object):
 
     @property
     def arxiv_categories(self):
-        categories = self.root.xpath('.//categories/text()').extract_first()
+        categories = self.root.xpath('.//categories/text()').extract_first(default='[]')
 
-        if categories:
-            return categories.split()
-
-        return []
+        return categories.split()
 
     @property
     def document_type(self):
         comments = '; '.join(self.root.xpath('.//comments/text()').extract())
-        if not comments:
-            return 'article'
         
         doctype = 'article'
         if RE_THESIS.search(comments):
--- a/hepcrawl/parsers/arxiv.py
+++ b/hepcrawl/parsers/arxiv.py
@@ -1,0 +1,367 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of hepcrawl.
+# Copyright (C) 2015, 2016, 2017 CERN.
+#
+# hepcrawl is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Extracts for various metadata formats"""
+
+from __future__ import absolute_import, division, print_function
+
+import itertools
+import re
+
+import six
+
+from inspire_schemas.api import LiteratureBuilder, ReferenceBuilder
+from inspire_schemas.utils import split_page_artid
+from inspire_utils.date import PartialDate
+from inspire_utils.helpers import maybe_int, remove_tags
+
+from ..mappings import CONFERENCE_WORDS, THESIS_WORDS
+from ..utils import (
+    coll_cleanforthe,
+    get_first,
+    get_node,
+    split_fullname,
+    ParsedItem,
+    strict_kwargs,
+)
+
+RE_CONFERENCE = re.compile(
+    r'\b(%s)\b' % '|'.join(
+        [re.escape(word) for word in CONFERENCE_WORDS]
+    ),
+    re.I | re.U,
+)
+RE_THESIS = re.compile(
+    r'\b(%s)\b' % '|'.join(
+        [re.escape(word) for word in THESIS_WORDS]
+    ),
+    re.I | re.U,
+)
+RE_PAGES = re.compile(r'(?i)(\d+)\s*pages?\b')
+
+
+class ArxivParser(object):
+    """Parser for the arXiv format.
+
+    It can be used directly by invoking the :func:`ArxivParser.parse` method, or be
+    subclassed to customize its behavior.
+
+    Args:
+        arxiv_record (Union[str, scrapy.selector.Selector]): the record in arXiv format to parse.
+        source (Optional[str]): if provided, sets the ``source`` everywhere in
+            the record. Otherwise, the source is extracted from the arXiv metadata.
+    """
+    def __init__(self, arxiv_record, source=None):
+        self.root = self.get_root_node(arxiv_record)
+        if not source:
+            source = 'arXiv'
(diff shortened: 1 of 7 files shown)


hepcrawl/parsers/arxiv.py
```diff
@@ -0,0 +1,361 @@
...
+            self.builder.add_report_number(rep_number, self.source)
+        self.builder.add_arxiv_eprint(self.arxiv_eprint, self.arxiv_categories)
+        self.builder.add_private_note(self.private_note)
+        self.builder.add_document_type(self.document_type)
+
+        return self.builder.record
+
+    @property
+    def preprint_date(self):
+        preprint_date_node = self.root.xpath('.//created/text()').extract_first()
```
michamos: you shouldn't call this `_node`, as it's not a node.

hepcrawl/parsers/arxiv.py
```diff
@@ -0,0 +1,361 @@
...
+
+    @property
+    def abstract(self):
+        abstract = self.root.xpath('.//abstract/text()').extract_first()
+
+        return abstract
+
+    @property
+    def authors(self):
+        authors, _, __ = _get_authors_and_collaborations(self.root)
```
michamos: * `_` can be reused, so you could do `authors, _, _ = ...`
* you're calling several times `_get_authors_and_collaborations`, which might be an expensive function. You can cache it by doing something like:
```python
@property
def authors_and_collaborations(self):
    if not hasatrr(self, '_authors_and_collaborations'):
        self._authors_and_collaborations = _get_authors_and_collaborations(self.root)
    return self._authors_and_collaborations
```
and then using that property everywhere.
Better yet, you can define somewhere a `cached_property` decorator that does this in general and would be used in place of `property` for properties reused several times.

hepcrawl/parsers/arxiv.py
```diff
@@ -0,0 +1,361 @@
...
+        _, __, doctype = _parse_comments_info(comments)
+
+        return doctype
+
+    @property
+    def source(self):
+        return 'arXiv'
+
+
+def _parse_comments_info(comments):
```
michamos: I would turn this into a `staticmethod`, as it's some internal thing used by the parser.

hepcrawl/parsers/arxiv.py
```diff
@@ -0,0 +1,361 @@
...
+        elif RE_CONFERENCE.search(comments):
+            doctype = 'conference paper'
+
+        return pages, notes, doctype
+
+    else:
+        return None, None, 'article'
+
+
+def _get_authors_and_collaborations(node):
```
michamos: same.

hepcrawl/parsers/arxiv.py
```diff
@@ -0,0 +1,361 @@
...
+                'affiliations': affiliations_without_collaborations
+            })
+    if warning_tags:
+        warning = 'WARNING: Colon in authors before %s: Check author list for collaboration names!' % ', '.join(warning_tags)
+    else:
+        warning = ''
+    return authors, collaborations, warning
+
+
+def _get_author_names_and_affiliations(author_node):
```
michamos: same

tests/functional/arxiv/fixtures/arxiv_expected.json
```diff
@@ -12,10 +12,10 @@
         ], 
         "authors": [
             {
-                "full_name": "Almeida, C.A.S."
+                "full_name": "Almeida, C.A. S."
```
michamos: how does this happen?
```
In [1]: from inspire_schemas.api import LiteratureBuilder

In [2]: l = LiteratureBuilder()

In [3]: l.make_author('Almeida, C.A. S.')
Out[3]: {'full_name': u'Almeida, C.A.S.'}
```

tests/functional/arxiv/fixtures/http_server/records/arxiv-physics-hep-ex.xml
```diff
@@ -10,18 +10,19 @@
...
-  Xenon Detectors</title><categories>physics.ins-det hep-ex</categories><comments>10 pages, 9 figures</comments><license>http://arxiv.org/licenses/nonexclusive-distrib/1.0/</license><abstract>  Silicon photomultipliers (SiPMs) are potential solid-state alternatives to
-traditional photomultiplier tubes (PMTs) for single-photon detection. In this
-paper, we report on evaluating SensL MicroFC-10035-SMT SiPMs for their
-suitability as PMT replacements. The devices were successfully operated in a
-liquid-xenon detector, which demonstrates that SiPMs can be used in noble
-element time projection chambers as photosensors. The devices were also cooled
-down to 170 K to observe dark count dependence on temperature. No dependencies
-on the direction of an applied 3.2 kV/cm electric field were observed with
-respect to dark-count rate, gain, or photon detection efficiency.
-</abstract></arXiv>
```
michamos: what is the response from arXiv? If there are new lines there, there should be newlines here, if not there shouldn't. You should not mess with the response in any case.

michamos:
> In the unit tests, this path has been proven to be `./metadata/arxiv/[FIELD]/text()` however, in the functional tests, it was `./[FIELD]/text()`, so the only way to make both work is to leave the paths as they previously were; `.//[FIELD]/text()`

That's messed up. Does it depend on `arXiv` vs `arXiv_single`? or do some of the tests do some pre-processing they shouldn't?

vbalbp:
> > In the unit tests, this path has been proven to be `./metadata/arxiv/[FIELD]/text()` however, in the functional tests, it was `./[FIELD]/text()`, so the only way to make both work is to leave the paths as they previously were; `.//[FIELD]/text()`
> 
> That's messed up. Does it depend on `arXiv` vs `arXiv_single`? or do some of the tests do some pre-processing they shouldn't?

It depends on pre-processing that they shouldn't. Both ``arXiv`` and ``arXiv_single`` work for unit tests and the both fail for functional tests if I set the xpath to ``./metadata/arxiv/[FIELD]/text()``.
[In the unit tests](https://github.com/inspirehep/hepcrawl/blob/823f41daf4f2be6d2ff17fcb2b598f726d7ab694/tests/unit/test_arxiv_all.py#L51), you can see that it's getting every record by using the xpath formula ``.//record``.
However, somewhere in the functional tests, it's parsing it in a different way, returning a list of ``metadata`` elements, which are elements that in the previous syntax, hung from ``record``. [These crawl results](https://github.com/inspirehep/hepcrawl/blob/823f41daf4f2be6d2ff17fcb2b598f726d7ab694/tests/functional/arxiv/test_arxiv.py#L113) already have the ``metadata`` elements mentioned before, so I was not able to figure out exactly where this pre-processing was being done.

tests/functional/arxiv/fixtures/arxiv_expected.json
vbalbp: > how does this happen?
> 
> ```
> In [1]: from inspire_schemas.api import LiteratureBuilder
> 
> In [2]: l = LiteratureBuilder()
> 
> In [3]: l.make_author('Almeida, C.A. S.')
> Out[3]: {'full_name': u'Almeida, C.A.S.'}
> ```

Like this:
```
In [1]: from inspire_schemas.api import LiteratureBuilder

In [2]: l = LiteratureBuilder()

In [3]: l.make_author('Almeida, C. A. S.')
Out[3]: {'full_name': u'Almeida, C.A. S.'}
```

tests/unit/responses/arxiv/sample_arxiv_record10.xml
```diff
@@ -10,34 +10,19 @@
  <setSpec>physics:hep-th</setSpec>
 </header>
 <metadata>
- <arXiv xmlns="http://arxiv.org/OAI/arXiv/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://arxiv.org/OAI/arXiv/ http://arxiv.org/OAI/arXiv.xsd">
- <id>1606.04259</id><created>2016-06-14</created><authors><author><keyname>Battista</keyname><forenames>Emmanuele</forenames></author></authors><title>Extreme Regimes in Quantum Gravity</title><categories>hep-th</categories><comments>Phd thesis in Fundamental and Applied Physics presented at University
-  &quot;Federico II&quot; (Naples, Italy) on 29th April 2016</comments><license>http://arxiv.org/licenses/nonexclusive-distrib/1.0/</license><abstract>  The thesis is divided into two parts. In the first part the low-energy limit
-of quantum gravity is analysed, whereas in the second we deal with the
```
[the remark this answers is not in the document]
vbalbp: Indeed it doesn't, I forgot to change it back when I was meddling with the test samples

vbalbp:
The response from arXiv is returned in the same way as the fixture is set, with random end-of-line characters in the middle of long strings like title, comments and abstract. ``fix_abstract`` and ``fix_title`` work as intended.

hepcrawl/parsers/arxiv.py
```diff
@@ -0,0 +1,363 @@
...
+
+        return rns
+    
+    @property
+    def arxiv_eprint(self):
+        return self.root.xpath('.//id/text()').extract_first()
+
+    @property
+    def arxiv_categories(self):
+        categories = self.root.xpath('.//categories/text()').extract_first()
```
michamos: you can pass a `default=''` to `extract_first`, then you don't have to test if `categories` is truthy, as `''.split() == []`.

hepcrawl/parsers/arxiv.py
```diff
@@ -0,0 +1,363 @@
...
+
+        if categories:
+            return categories.split()
+
+        return []
+
+    @property
+    def document_type(self):
+        comments = '; '.join(self.root.xpath('.//comments/text()').extract())
+        if not comments:
```
michamos: is this really needed? `comments` is always a string because of the `join`, so in the worst case the regex searches find nothing and you return `'article'` at the end.

PR #257, opened 2019-01-07
oaipmh: gracefully parse records

drjova:
Signed-off-by: Harris Tzovanakis <me@drjova.com>


hepcrawl/spiders/common/oaipmh_spider.py
```diff
@@ -227,7 +227,11 @@ def parse_list(self, response):
             self._crawled_records[rec_identifier] = record
             response = XmlResponse(self.url, encoding='utf-8', body=record.raw)
             selector = Selector(response, type='xml')
-            yield self.parse_record(selector)
+
+            try:
+                yield self.parse_record(selector)
+            except Exception:
+                LOGGER.warning('Skipping, cannot parse record %s', rec_identifier)
```
michamos: maybe log the exception too to have a bit more info? I would also log an error rather than a warning.

drjova:
updated

PR #261, opened 2019-02-06
APS spider: use LastRunStore spider

vbalbp:

### DIFF
--- a/hepcrawl/spiders/aps_spider.py
+++ b/hepcrawl/spiders/aps_spider.py
@@ -14,13 +14,14 @@ from __future__ import absolute_import, division, print_function
 import json
 import link_header
 
+from datetime import datetime
 from furl import furl
-
 from scrapy import Request
 
 from inspire_utils.record import get_value
 
 from . import StatefulSpider
+from .common.lastrunstore_spider import LastRunStoreSpider
 from ..items import HEPRecord
 from ..loaders import HEPLoader
 from ..parsers import JatsParser
@@ -32,7 +33,7 @@ from ..utils import (
 )
 
 
-class APSSpider(StatefulSpider):
+class APSSpider(LastRunStoreSpider):
     """APS crawler.
 
     Uses the APS REST API v2.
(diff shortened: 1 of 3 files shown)


hepcrawl/spiders/aps_spider.py
```diff
@@ -199,3 +213,6 @@ def _get_authors_and_collab(self, article):
 
     def _file_name_from_url(self, url):
         return "{}.xml".format(url[url.rfind('/') + 1:])
+
+    def make_file_fingerprint(self, set_):
+        return u'metadataPrefix={}&set={}'.format(self.format, set_)
```
michamos: where is the `format` coming from? do you really need that part? it seems OAI-PMH specific. Are you sure this really works?

hepcrawl/spiders/aps_spider.py
```diff
@@ -41,43 +42,48 @@ class APSSpider(StatefulSpider):
     Example:
         Using the APS spider::
 
-            $ scrapy crawl APS -a 'from_date=2016-05-01' -a 'until_date=2016-05-15' -a 'set=openaccess'
+            $ scrapy crawl APS -a 'from_date=2016-05-01' -a 'until_date=2016-05-15' -a 'sets=openaccess'
 
     .. _See documentation here:
         http://harvest.aps.org/docs/harvest-api#endpoints
+
+    Journals are not supported as a parameter anymore for APS Spider
```
michamos:
```
Note:
    Selecting specific journals is not supported for technical reasons as it's incompatible with the way the last run time is stored.
```

hepcrawl/spiders/aps_spider.py
```diff
@@ -41,43 +42,47 @@ class APSSpider(StatefulSpider):
...
+    @property
+    def url(self):
+        params = {
+            'sets': self.set,
+            'from_date': self.from_date or self.resume_from(set_=self._set),
+            'until_date': self.until_date,
+            'per_page': self.per_page,
+            'date': self.date
+        }
+        return furl(APSSpider.aps_base_url).add(params).url
```
michamos: are the `None` checks that were done previously useless?

vbalbp: I tried it out and it does matter, so changing it back to the old way with the checks

PR #266, opened 2019-03-22
tests: python2/3 compatibility

tsgit:
* ensure tests pass under python2 and python3

    * use deepdiff for comparing nested data-structures

    * remove deep_sort, because sorting a list of dicts
      is not supported in python3

    * accommodate differences in str types between python 2/3

Signed-off-by: Thorsten Schwander <thorsten.schwander@gmail.com>

## Description

improve compatibility with python3 by using six
update tests so that all existing unit tests pass with both python 2.7 and python 3.7

## Checklist:

### DIFF
--- a/hepcrawl/spiders/elsevier_spider.py
+++ b/hepcrawl/spiders/elsevier_spider.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of hepcrawl.
-# Copyright (C) 2016, 2017 CERN.
+# Copyright (C) 2016, 2017, 2019 CERN.
 #
 # hepcrawl is a free software; you can redistribute it and/or modify it
 # under the terms of the Revised BSD License; see LICENSE file for
@@ -13,29 +13,27 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import re
-
 from tempfile import mkdtemp
 
 import dateutil.parser as dparser
-
 import requests
-
 from scrapy import Request
 from scrapy.spiders import XMLFeedSpider
+from six import text_type
 
 from . import StatefulSpider
+from ..dateutils import format_year
 from ..items import HEPRecord
 from ..loaders import HEPLoader
 from ..utils import (
+    ParsedItem,
     get_first,
     get_licenses,
     has_numbers,
     range_as_string,
-    unzip_xml_files,
-    ParsedItem,
     strict_kwargs,
+    unzip_xml_files,
 )
-from ..dateutils import format_year
 
 
 class ElsevierSpider(StatefulSpider, XMLFeedSpider):
@@ -355,7 +353,7 @@ class ElsevierSpider(StatefulSpider, XMLFeedSpider):
 
         # return year, raw_date
         # NOTE: I have call this here (not in the loader), because I need the year.
-        return format_year(raw_date), unicode(raw_date)
+        return format_year(raw_date), text_type(raw_date)
 
     def get_doctype(self, node):
         """Return a doctype mapped from abbreviation."""
@@ -471,7 +469,7 @@ class ElsevierSpider(StatefulSpider, XMLFeedSpider):
         elif trans_title:  # NOTE: is this case even possible?
             title = trans_title
 
-        return unicode(title)
+        return text_type(title)
 
     @staticmethod
     def _get_ref_journal_title(ref):
@@ -713,7 +711,7 @@ class ElsevierSpider(StatefulSpider, XMLFeedSpider):
             basename = os.path.basename(xml_file)
             elsevier_id = os.path.splitext(basename)[0]
             url = u"http://www.sciencedirect.com/science/article/pii/" + elsevier_id
-        except AttributeError:
+        except (AttributeError, TypeError):
             url = ''
         return url
 
@@ -739,7 +737,7 @@ class ElsevierSpider(StatefulSpider, XMLFeedSpider):
 
         try:
             # filter after re.split, which may return empty elements:
-            split_pub = filter(None, re.split(r'(\W+)', publication))
+            split_pub = [element for element in re.split(r'(\W+)', publication) if element]
             if split_pub[-1] in possible_sections:
                 section = split_pub.pop(-1)
 
--- a/hepcrawl/spiders/pos_spider.py
+++ b/hepcrawl/spiders/pos_spider.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of hepcrawl.
-# Copyright (C) 2016, 2017 CERN.
+# Copyright (C) 2016, 2017, 2019 CERN.
 #
 # hepcrawl is a free software; you can redistribute it and/or modify it
 # under the terms of the Revised BSD License; see LICENSE file for
@@ -11,21 +11,21 @@
 
 from __future__ import absolute_import, division, print_function
 
-import re
 import os
-from urlparse import urljoin
-from six.moves.urllib.parse import quote
+import re
 
 from scrapy import Request, Selector
+
+from six.moves.urllib.parse import quote, urljoin
 
 from . import StatefulSpider
 from ..dateutils import create_valid_date
 from ..items import HEPRecord
 from ..loaders import HEPLoader
 from ..utils import (
-    get_licenses,
-    get_first,
     ParsedItem,
+    get_first,
+    get_licenses,
     strict_kwargs,
 )
 
--- a/hepcrawl/spiders/wsp_spider.py
+++ b/hepcrawl/spiders/wsp_spider.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of hepcrawl.
-# Copyright (C) 2015, 2016, 2017 CERN.
+# Copyright (C) 2015, 2016, 2017, 2019 CERN.
 #
 # hepcrawl is a free software; you can redistribute it and/or modify it
 # under the terms of the Revised BSD License; see LICENSE file for
@@ -12,21 +12,22 @@
 from __future__ import absolute_import, division, print_function
 
 import os
-import urlparse
 import tempfile
 
 from scrapy import Request
 from scrapy.spiders import XMLFeedSpider
 
+from six.moves.urllib.parse import urlsplit
+
 from . import StatefulSpider
 from ..parsers import JatsParser
 from ..utils import (
-    ftp_list_files,
-    ftp_connection_info,
-    local_list_files,
-    unzip_xml_files,
     ParsedItem,
+    ftp_connection_info,
+    ftp_list_files,
+    local_list_files,
     strict_kwargs,
+    unzip_xml_files,
 )
 
 
@@ -212,7 +213,7 @@ class WorldScientificSpider(StatefulSpider, XMLFeedSpider):
     def handle_package_file(self, response):
         """Handle a local zip package and yield every XML."""
         self.logger.info("Visited file %s" % response.url)
-        zip_filepath = urlparse.urlsplit(response.url).path
+        zip_filepath = urlsplit(response.url).path
         xml_files = unzip_xml_files(zip_filepath, self.destination_folder)
 
         for xml_file in xml_files:
--- a/hepcrawl/utils.py
+++ b/hepcrawl/utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of hepcrawl.
-# Copyright (C) 2015, 2016, 2017, 2018 CERN.
+# Copyright (C) 2015, 2016, 2017, 2018, 2019 CERN.
 #
 # hepcrawl is a free software; you can redistribute it and/or modify it
 # under the terms of the Revised BSD License; see LICENSE file for
@@ -20,16 +20,16 @@ from operator import itemgetter
 from itertools import groupby
 from netrc import netrc
 from zipfile import ZipFile
-from urlparse import urlparse
 
 import ftputil
 import ftputil.session
 import ftplib
-from hepcrawl.tohep import hep_to_hep, _normalize_hepcrawl_record, \
-    hepcrawl_to_hep, UnknownItemFormat
 from inspire_schemas.builders import LiteratureBuilder
-
 from scrapy import Selector
+
+from hepcrawl.tohep import (hep_to_hep, _normalize_hepcrawl_record,
+                            hepcrawl_to_hep, UnknownItemFormat)
+from six.moves.urllib.parse import urlparse
 
 RE_FOR_THE = re.compile(
     r'\b(?:for|on behalf of|representing)\b',
@@ -238,9 +238,9 @@ def range_as_string(data):
     ranges = []
     for key, group in groupby(
         enumerate(data),
-        lambda (index, item): index - item
+        lambda index_item: index_item[0] - index_item[1]
     ):
-        group = map(itemgetter(1), group)
+        group = [year for _, year in group]
         if len(group) > 1:
             rangestring = "{}-{}".format(str(group[0]), str(group[-1]))
             ranges.append(rangestring)
@@ -289,7 +289,7 @@ def get_journal_and_section(publication):
     journal_title = ''
     possible_sections = ["A", "B", "C", "D", "E"]
     try:
-        split_pub = filter(None, re.split(r'(\W+)', publication))
+        split_pub = [element for element in re.split(r'(\W+)', publication) if element]
         if split_pub[-1] in possible_sections:
             section = split_pub.pop(-1)
         journal_title = "".join(
--- a/tests/functional/arxiv/test_arxiv.py
+++ b/tests/functional/arxiv/test_arxiv.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of hepcrawl.
-# Copyright (C) 2017 CERN.
+# Copyright (C) 2017, 2019 CERN.
 #
 # hepcrawl is a free software; you can redistribute it and/or modify it
 # under the terms of the Revised BSD License; see LICENSE file for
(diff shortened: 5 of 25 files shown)


hepcrawl/spiders/elsevier_spider.py
```diff
@@ -739,7 +740,7 @@ def get_journal_and_section(publication):
 
         try:
             # filter after re.split, which may return empty elements:
-            split_pub = filter(None, re.split(r'(\W+)', publication))
+            split_pub = list(filter(None, re.split(r'(\W+)', publication)))
```
michamos: i would transform this into a list comprehension, I find `filter(None,...)` very non-obvious:
```suggestion
            split_pub = [element in re.split(r'(\W+)', publication)) if element]
```

hepcrawl/spiders/pos_spider.py
```diff
@@ -13,7 +13,7 @@
 
 import re
 import os
-from urlparse import urljoin
+from six.moves.urllib.parse import urljoin
 from six.moves.urllib.parse import quote
```
michamos: you can combine these two imports

hepcrawl/spiders/wsp_spider.py
```diff
@@ -12,7 +12,7 @@
 from __future__ import absolute_import, division, print_function
 
 import os
-import urlparse
+from six.moves.urllib.parse import urlsplit
```
michamos: `from ...` imports should go after normal module `import`s (so one line lower)

hepcrawl/utils.py
```diff
@@ -20,7 +20,8 @@
 from itertools import groupby
 from netrc import netrc
 from zipfile import ZipFile
-from urlparse import urlparse
+from six.moves.urllib.parse import urlparse
+# from urlparse import urlparse
```
michamos: leftover comment?

hepcrawl/utils.py
```diff
@@ -238,9 +239,9 @@ def range_as_string(data):
     ranges = []
     for key, group in groupby(
         enumerate(data),
-        lambda (index, item): index - item
+            lambda index_item: index_item[0] - index_item[1]
```
michamos: sad that you can't unpack tuples in `lambda`s in Python 3. what's with the strange indentation though?

hepcrawl/utils.py
```diff
@@ -238,9 +239,9 @@ def range_as_string(data):
...
     ):
-        group = map(itemgetter(1), group)
+        group = list(map(itemgetter(1), group))
```
michamos:
```suggestion
        group = [year for (_, year) in group]
```
(not that it matters too much, I have no idea why such a function doing formatting would be needed in a software that *extracts* metadata).

hepcrawl/utils.py
```diff
@@ -289,7 +290,7 @@ def get_journal_and_section(publication):
     journal_title = ''
     possible_sections = ["A", "B", "C", "D", "E"]
     try:
-        split_pub = filter(None, re.split(r'(\W+)', publication))
+        split_pub = list(filter(None, re.split(r'(\W+)', publication)))
```
michamos: I'm having some [déjà vu](https://github.com/inspirehep/hepcrawl/pull/266/files#diff-a72568f5025a10728db1499d3e6df627R743) here :)

tests/functional/arxiv/test_arxiv.py
```diff
@@ -122,8 +123,5 @@ def test_arxiv(
         override_generated_fields(expected) for expected in expected_results
     ]
 
-    gotten_results = deep_sort(gotten_results)
-    expected_results = deep_sort(expected_results)
-
-    assert gotten_results == expected_results
+    assert DeepDiff(gotten_results, expected_results, ignore_order=True) == {}
```
michamos: I'm not sure whether this is actually what we want, as ordering can be meaningful. The correct thing here is probably to do a shallow sort: we don't care in which order the records arrive (I see no reason why that would be non-deterministic in a test environment though), but we do care that the records are really equal, not equal up to arbitrary permutations, of say, the authors on the paper, as that's a sign of bugs :cry: not something I expect you to fix in this PR though.

tests/unit/test_desy.py
```diff
@@ -111,23 +114,29 @@ def test_pipeline(generated_records, expected_records):
...
     with open(path, 'r') as xmlfile:
         data = xmlfile.read()
     result = spider._parsed_items_from_marcxml([data])
-    expected_exception = (
-        "DoJsonError(u'Error in rule \"imprints\" for field \"260__\"',"
-        " (u'Unknown string format:', 'Not a date'), GroupableOrderedDict((("
-        "'a', 'Hamburg'), ('c', 'Not a date'), "
-        "('b', 'Verlag Deutsches Elektronen-Synchrotron'))))"
-    )
+    if sys.version_info[0] < 3:
```
michamos: can you weaken the test instead? this is testing that exceptions inside the spider due to bad MARC:
1. don't cause the spider to blow up
2. are correctly recorded

So I would do something simple like this instead:
```python
assert result[0].exception.startswith('DoJsonError')
```

tests/unit/test_edp.py
```diff
@@ -57,7 +55,7 @@ def package_jats(targzfile):
     """Extract tar.gz package with JATS XML file."""
     spider = edp_spider.EDPSpider()
     response = fake_response_from_string(text="", url="file://" + targzfile)
-    return spider.handle_package_file(response).next()
+    return six.next(spider.handle_package_file(response))
```
michamos: you don't need `six` for this if you only need compatibility with Python 2.7+, 
```suggestion
    return next(spider.handle_package_file(response))
```

tests/unit/test_edp.py
```diff
@@ -100,7 +98,7 @@ def package_rich(tarbzfile):
     """Extract tar.gz package with 'rich' XML file."""
     spider = edp_spider.EDPSpider()
     response = fake_response_from_string(text="", url="file://" + tarbzfile)
-    return spider.handle_package_file(response).next()
+    return six.next(spider.handle_package_file(response))
```
michamos:
```suggestion
    return next(spider.handle_package_file(response))
```

tests/unit/test_edp.py
```diff
@@ -359,7 +357,7 @@ def test_handle_package_ftp(tarbzfile):
     """Test getting the target folder name for xml files."""
     spider = edp_spider.EDPSpider()
     response = fake_response_from_string(text=tarbzfile)
-    request = spider.handle_package_ftp(response).next()
+    request = six.next(spider.handle_package_ftp(response))
```
michamos:
```suggestion
    request = next(spider.handle_package_ftp(response))
```

tests/unit/test_parsers_jats.py
```diff
@@ -96,7 +98,10 @@ def test_field(field_name, records):
     result = getattr(records['jats'], field_name)
     expected = records['expected'][field_name]
 
-    assert result == expected
+    diffs = DeepDiff(result, expected, ignore_order=True)
+    if sys.version_info[0] < 3 and 'type_changes' in diffs:
+        del diffs['type_changes']
```
michamos: do you really need `DeepDiff` here? if nothing but the string type has changed, there won't be an issue, as
`"foo" == u"foo"` (with ASCII only).

tests/unit/test_pipelines.py
```diff
@@ -48,7 +49,7 @@ def json_spider_record(tmpdir):
 
     test_selectors = fake_response.xpath('.//record')
     items = (spider.parse_record(sel) for sel in test_selectors)
-    parsed_record = items.next()
+    parsed_record = six.next(items)
```
michamos:
```suggestion
    parsed_record = next(items)
```

tests/unit/test_pos.py
```diff
@@ -53,11 +54,11 @@ def generated_conference_paper(scrape_pos_conference_paper_page_body):
 
     crawler = Crawler(spidercls=pos_spider.POSSpider)
     spider = pos_spider.POSSpider.from_crawler(crawler)
-    request = spider.parse(
+    request = six.next(spider.parse(
```
michamos:
```suggestion
    request = next(spider.parse(
```

tests/unit/test_pos.py
```diff
@@ -68,7 +69,7 @@ def generated_conference_paper(scrape_pos_conference_paper_page_body):
 
     pipeline = InspireCeleryPushPipeline()
     pipeline.open_spider(spider)
-    parsed_item = request.callback(response).next()
+    parsed_item = six.next(request.callback(response))
```
michamos:
```suggestion
    parsed_item = next(request.callback(response))
```

tests/unit/test_world_scientific.py
```diff
@@ -66,7 +67,7 @@ def get_records(response_file_name):
 
 def get_one_record(response_file_name):
     records = get_records(response_file_name)
-    record = records.next()
+    record = six.next(records)
```
michamos:
```suggestion
    record = next(records)
```

tests/unit/test_parsers_jats.py
tsgit: the expected output is read from `YAML`. yaml.load() only marks non-ascii as `u''`
the `jats` parser marks everything as `u''`.

for string comparison this doesn't matter, but here we compare nested data structure with lists of dicts

I use DeepDiff to compare records with `ignore_order=True` because there are lists of dicts (e.g. affiliations) which do not compare as equal unless they are in the same order, however in python3 you can't sort a list of dicts easily.

```python3
Python 3.7.2 (default, Mar 21 2019, 10:09:12) 
>>> sorted([{2:3,4:5}, {3:1, 1:2}])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'dict' and 'dict'
>>> 
```
```python2
Python 2.7.15 (default, Oct 15 2018, 15:26:09) 
>>> sorted([{2:3,4:5}, {3:1, 1:2}])
[{1: 2, 3: 1}, {2: 3, 4: 5}]
```

DeepDiff does do a type comparison, and in python2 the types are different

```python2
>>> type('hello')
<type 'str'>
>>> type(u'hello')
<type 'unicode'>
```

while in python3 
```
>>> type('hello')
<class 'str'>
>>> type(u'hello')
<class 'str'>
```

DeepDiff has a `ignore-type` option in current version, but the last version compatible with python2.7 doesn't have that option.

tests/unit/test_edp.py
tsgit: right, I see now  "The next() built-in was backported to Python 2.6+, where it calls the next method."

hepcrawl/utils.py
tsgit: I agree that comprehensions are more readable than `list(map(  ))`  constructs. 
I should have taken the opportunity to re-write this and other similar expressions
I think `[g[1] for g in group]` is better, because it doesn't rely on group having exactly 2 values to unpack

tests/functional/arxiv/test_arxiv.py
tsgit: mmh, good point. I assumed that `deep_sort` was done with purpose and therefore simply looked at an equivalent way of comparing records. I do get different test results depending on what venv and what python version I run in, so the field order is not stable. certainly author lists should remain in stable sort order, but what about e.g. keywords etc.
anyhow, with `DeepDiff` you can set option `ignore_order=False` and have a direct comparison, but then that should be equivalent to `dict1 == dict2` 
I see if I can find reproducible examples.

tsgit:
I agree that `python: - '3.7'` should be added to .travis.yml .
I'm not sure it'll work without some other modifications, though

I use DeepDiff everywhere a direct comparison `expected = result` fails otherwise with current tests. those tests should probably be improved instead.

tsgit:
mmh, so for example in the desy test the MARC tags in the XML file aren't sorted numerically 
which should be fine, since that's not a requirement

however the sort order of the expected json reflects that -- while I think the author from MARC 100 should be before MARC 70x in the converted json

https://github.com/inspirehep/hepcrawl/blob/master/tests/unit/responses/desy/desy_record.xml
`701` appears before `100`

https://github.com/inspirehep/hepcrawl/blob/master/tests/unit/responses/desy/desy_record_expected.json
preserves that order

the actual test output is correct in this case and changes the order to what it should be

```
 'authors': [{'affiliations': [{'value': 'DESY'}],
   'full_name': 'Turkot, Oleksii',
   'ids': [{'schema': 'ORCID', 'value': '0000-0001-5352-7744'}],
   'raw_affiliations': [{'value': 'Deutsches Elektronen-Synchrotron'}]},
  {'full_name': 'Foster, Brian',
   'ids': [{'schema': 'ORCID', 'value': '0000-0001-5699-3046'}],
   'inspire_roles': ['supervisor'],
   'raw_affiliations': [{'value': 'Deutsches Elektronen-Synchrotron'}]},
  {'full_name': 'Wichmann, Katarzyna',
   'inspire_roles': ['supervisor'],
   'raw_affiliations': [{'value': 'Deutsches Elektronen-Synchrotron'}]}],
```

so I guess `deep_sort` was a flawed and unnecessary workaround ?

tests/unit/test_parsers_jats.py
tsgit: @michamos 

after changing this to `assert expected == result` I get occasional test errors on subsequent runs -- seemingly depending on some external state of machine 

`$ pytest -vv tests/unit/test_parsers_jats.py`

it doesn't help to do `rm -rf .pytest_cache/`

```
field_name = 'authors'
records = {'expected': {'abstract': 'The study of non-Abelian Majorana zero modes advances our understanding of the fundamental ...'}]}], ...}, 'file_name': 'PhysRevX.4.021018.xml', 'jats': <hepcrawl.parsers.jats.JatsParser object at 0x7ffb2dc08390>}

    @pytest.mark.parametrize(
        'field_name',
        FIELDS_TO_CHECK
    )
    def test_field(field_name, records):
        result = getattr(records['jats'], field_name)
        expected = records['expected'][field_name]
    
>       assert expected == result
E       AssertionError: assert [{'full_name'...ng, China'}]}] == [{'full_name':...ng, China'}]}]
E         At index 0 diff: {'raw_affiliations': [{'source': 'American Physical Society', 'value': 'Department of Physics, Hong Kong University of Science and Technology, Clear Water Bay, Hong Kong, China'}, {'source': 'American Physical Society', 'value': 'Institute for Advanced Study, Hong Kong University of Science and Technology, Clear Water Bay, Hong Kong, China'}], 'full_name': 'Liu, Xiong-Jun'} != {'full_name': 'Liu, Xiong-Jun', 'raw_affiliations': [{'value': 'Institute for Advanced Study, Hong Kong University of Science and Technology, Clear Water Bay, Hong Kong, China', 'source': 'American Physical Society'}, {'value': 'Department of Physics, Hong Kong University of Science and Technology, Clear Water Bay, Hong Kong, China', 'source': 'American Physical Society'}]}
E         Full diff:
E         [{'full_name': 'Liu, Xiong-Jun',
E         'raw_affiliations': [{'source': 'American Physical Society',
E         +                         'value': 'Institute for Advanced Study, Hong Kong '
E         +                                  'University of Science and Technology, Clear '
E         +                                  'Water Bay, Hong Kong, China'},
E         +                        {'source': 'American Physical Society',
E         'value': 'Department of Physics, Hong Kong University '
E         'of Science and Technology, Clear Water Bay, '
E         -                                  'Hong Kong, China'},
E         +                                  'Hong Kong, China'}]},
E         ?                                                     ++
E         -                        {'source': 'American Physical Society',
E         -                         'value': 'Institute for Advanced Study, Hong Kong '
E         -                                  'University of Science and Technology, Clear '
E         -                                  'Water Bay, Hong Kong, China'}]},
E         {'full_name': 'Wong, Chris L.M.',
E         'raw_affiliations': [{'source': 'American Physical Society',
E         'value': 'Department of Physics, Hong Kong University '
E         'of Science and Technology, Clear Water Bay, '
E         'Hong Kong, China'}]},
E         {'full_name': 'Law, K.T.',
E         'raw_affiliations': [{'source': 'American Physical Society',
E         'value': 'Department of Physics, Hong Kong University '
E         'of Science and Technology, Clear Water Bay, '
E         'Hong Kong, China'}]}]

tests/unit/test_parsers_jats.py:101: AssertionError
... 
====================================== 1 failed, 74 passed, 3 skipped, 2 warnings in 18.99 seconds ======================================
```
 
it is always the author field that errors

```
for x in `seq 1 10`; do pytest tests/unit/test_parsers_jats.py ; done
```
gives  test pass/fail   `+ - + - + + + + - -`  so there is no discernible pattern

trying with cache removal:
```
for x in `seq 1 10; do rm -rf .pytest_cache; pytest tests/unit/test_parsers_jats.py  |  grep seconds; done
```
`+ + + - - - + - - -`

typically the state changes when I do a bit of mouse movement !

T.

tsgit: so I suggest to make clear the `authors` test is special

```
    if field_name == 'authors':
        diffs = DeepDiff(result, expected, ignore_order=True)
        if sys.version_info[0] < 3 and 'type_changes' in diffs:
            del diffs['type_changes']
        assert diffs == {}
    else:
        assert result == expected
```
 and leave it to a future improvement to figure out stable sort order

michamos:
@tsgit on python 3 the documentation fails to build. This seems to be due to it using Sphinx 2.0 (which is Python 3 only) and the hepcrawl config being incompatible with it. You can either try fixing the config to make it compatible, or if you don't wan to bother simply pin sphinx to `<2.0`.

tsgit:
well there is a snowball effect of things just in building the docs

I fixed the immediate issue with sphinx
however there is also an issue with the use of autosemver and python 3
and then there is an issue with scrapyd < 2.0 and python3
and maybe other things

michamos:
@tsgit It would probably be easiest to disable docs building in https://github.com/inspirehep/hepcrawl/blob/02a2d9919c8b1ce904c1b7a85d1072e2ba87834e/docker-compose.test.py3.yml#L69. What do you think?

PR #267, opened 2019-03-28
travis: also test on Python 3

michamos:
This requires a new Docker image running Python 3. The simplest (but a
bit dirty) way to test on both Python 2 and Python 3 is to duplicate the
`docker-compose` YAMLs.

As the Python 3 tests currently fail, failure is allowed.


tests/Dockerfile.hepcrawl_py3
```diff
@@ -0,0 +1,11 @@
+FROM python:3-slim
+
+WORKDIR /usr/src/app
+
+RUN apt-get update \
+  && apt-get install --no-install-recommends -y build-essential \
+  && rm -rf /var/lib/apt/lists
+
+COPY . .
```
ammirate: Since you are copying everything, you might want to define a `.dockerignore` file avoiding copying tests and other useless files.

docker-compose.test.py3.yml
```diff
@@ -0,0 +1,201 @@
...
+    depends_on:
+      scrapyd:
+        condition: service_healthy
+      http-server.local:
+        condition: service_healthy
+
+  unit:
+    <<: *service_base
+    command: bash -c "py.test tests/unit -vv && make -C docs clean && make -C docs html && python setup.py sdist && ls dist/*"
+    links: []
```
ammirate: Is this needed?

docker-compose.test.py3.yml
```diff
@@ -0,0 +1,201 @@
...
+    depends_on:
+      celery:
+        condition: service_started
+    healthcheck:
+      timeout: 5s
+      interval: 5s
+      retries: 5
+      test:
+        - "CMD-SHELL"
+        - "curl http://localhost:6800/listprojects.json"
```
ammirate: Might be useful to check something in the response to be sure everything is ok.

ammirate: Eventually, also in other healtchecks :point_down:

PR #280, opened 2019-10-03
setup: Downgrade minimal celery version to 4.1

pazembrz:
* Downgrade minimum version of celery to 4.1 as 4.2 is not supported by inspire-next

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ install_requires = [
     'scrapyd-client>=1.0.1',
     'six>=1.9.0',
     'requests~=2.22,>=2.22.0',
-    'celery~=4.2',
+    'celery>=4.1',
     'redis>=2.10.5',
     'pyasn1>=0.1.8',  # Needed for dependency resolving.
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -30,7 +30,7 @@
     'scrapyd-client>=1.0.1',
     'six>=1.9.0',
     'requests~=2.22,>=2.22.0',
-    'celery~=4.2',
+    'celery>=4.1',
```
michamos: don't we use 4.2 on hep, which also depends on hepcrawl? why not bump celery in next instead?

pazembrz: On hep we use 4.2 on next we are locked on 4.1 ad 4.2 gives us `DecodeError: No JSON object could be decoded` for error thrown in worker: `signal only works in main thread` This is old thing we decided that we wont' fix it for now.

michamos: yeah, sorry, I misread the `>=` for a `~=`.

PR #283, opened 2020-08-26
parsers: add elsevier parser

MJedr:
Ref: inspirehep/inspirehep#1208


hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,578 @@
...
+
+        Args:
+            ref_node(scrapy.selector.Selector): a selector on a single reference.
+
+        Returns:
+            List[str]: list of names
+        """
+        given_names = ref_node.xpath(".//contribution/authors/author/given-name/text()").extract()
+        last_names = ref_node.xpath(".//contribution/authors/author/surname/text()").extract()
+        return [" ".join([name, last_name]) for name, last_name in zip(given_names, last_names)]
```
michamos, line 478: are you sure you always have exactly one `given-name` and `surname` for each author? maybe it would be better to select `.//contribution/authors/author/`, then iterate on the nodes to extract given and surname for each of them.

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,578 @@
...
+
+        Args:
+            ref_node(scrapy.selector.Selector): a selector on a single reference.
+
+        Returns:
+            List[str]: list of names
+        """
+        given_names = ref_node.xpath(".//editors/authors/author/given-name/text()").extract()
+        last_names = ref_node.xpath(".//editors/authors/author/last-name/text()").extract()
+        return [" ".join([name, last_name]) for name, last_name in zip(given_names, last_names)]
```
michamos, line 492: same comment. Are you sure this is called `last-name` now as opposed to `surname` for authors?

michamos (commented):
a first round of comments, I didn't look in detail at the data yet.

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,578 @@
...
+        if abbrv_doctype:
+            doctype = DOCTYPE_MAPPING[get_first(abbrv_doctype)]
+        elif self.root.xpath(".//ja:article"):
+            doctype = "article"
+        elif self.root.xpath(".//ja:simple-article"):
+            doctype = "article"
+        elif self.root.xpath(".//ja:book-review"):
+            doctype = "book-review"
+        elif self.root.xpath(".//ja:exam"):
+            doctype = "exam"
```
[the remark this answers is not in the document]
MJedr: These are the doctypes of JA DTD. 
`The element exam is one of the top-level elements (doctypes) of the JA DTD`
(the same is for the book-review and simple-article).

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,578 @@
...
+        if self.root.xpath(".//conference-info"):
+            doctype = "conference_paper"
```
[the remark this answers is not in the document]
MJedr: No, the assumption was that if the document contains the node ```conference-info``` what might happen for ```Serial issue``` it is classified as the conference paper

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,578 @@
...
+            ce_affiliation = author_group.xpath(
+                "//affiliation[@id='" + aff_id + "']")
+            if ce_affiliation.xpath(".//affiliation"):
+                aff = ce_affiliation.xpath(
+                    ".//*[self::organization or self::city or self::country]/text()")
+                affiliations_by_id.append(", ".join(aff.extract()))
+            elif ce_affiliation:
+                aff = ce_affiliation.xpath(
+                    "./textfn/text()").extract_first()
+                aff = re.sub(r'^(\d+\ ?)', "", aff)
```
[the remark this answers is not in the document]
MJedr: These are for example postal codes of the institution, I don't know if it's needed

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,578 @@
...
+                          '|self::maintitle'
+                          '|self::volume-nr'
+                          '|self::issue-nr'
+                          '|self::date'
+                          '|self::label'
+                          '|self::publisher'
+                          '|self::object-id'
+                          '|self::issn'
+                          '|self:: doi'
+                          '|self::pages'
```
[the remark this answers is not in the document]
MJedr: This fields were removed in JATS parser, so I assumed that we do the same for Elsevier and I removed corresponding ones from this parser (especially those which are present in the other fields). I can add the comment for this less obvious ones.

hepcrawl/parsers/elsevier.py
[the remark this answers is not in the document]
michamos: you should keep everything (it should go to `raw_affiliations`, and we try to normalize it at a different stage)

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,555 @@
...
+
+    remove_tags_config_abstract = {
+        'allowed_tags': ['sup', 'sub'],
+        'allowed_trees': ['math'],
+        'strip': 'self::pub-id|self::issn'
+    }
+
+    @property
+    def abstract(self):
+        abstract_nodes = self.root.xpath(".//abstract-sec/simple-para")
```
michamos, line 133: can you try to make the selectors more specific, avoiding `//` directly under the root node? this has two effects:
* you won't match wrong things by accident (e.g. metadata for the references in case it's missing from the main document).
* it's more performant, as you don't have to traverse the whole tree to check whether nodes match.

tests/unit/responses/elsevier/j.nima.2019.162728.xml
```diff
@@ -0,0 +1 @@
...hors></sb:contribution><sb:host><sb:issue><sb:series><sb:title><sb:maintitle>EPJ Web Conf.</sb:maintitle></sb:title><sb:volume-nr>204</sb:volume-nr></sb:series><sb:date>2019</sb:date></sb:issue><sb:pages><sb:first-page>07007</sb:first-page></sb:pages><ce:doi>10.1051/epjconf/201920407007</ce:doi></sb:host></sb:reference><ce:source-text id="afs11">F. Guber et al., submitted to EPJ Web Conf. www.epj-conferences.org</ce:source-text></ce:bib-reference><ce:bib-reference id="b12"><ce:label>[12]</ce:label><ce:other-ref id="sb12"><ce:textref><ce:inter-ref id="interref5" xlink:href="https://ieeexplore.ieee.org/document/5402172" xlink:type="simple">https://ieeexplore.ieee.org/document/5402172</ce:inter-ref>.</ce:textref></ce:other-ref></ce:bib-reference><ce:bib-reference id="b13"><ce:label>[13]</ce:label><ce:other-ref id="sb13"><ce:textref>F. Guber, et al. <ce:inter-ref id="interref6" xlink:href="http://repository.gsi.de/record/109059/files/20150720-CBM-TDR-PSD.pdf" xlink:type="simple">http://repository.gsi.de/record/109059/files/20150720-CBM-TDR-PSD.pdf</ce:inter-ref>.</ce:textref></ce:other-ref></ce:bib-reference></ce:bibliography-sec></ce:bibliography></ja:tail></ja:article></doc:document>
```
michamos, line 1: I don't think we're allowed to put this in a public repo. The main problem is the full content of the article being available. Can you replace the full-text by some dummy text (like `Full-text removed for copyright reasons`), keeping the structure?

tests/unit/responses/elsevier/j.nima.2019.162728_expected.yml
```diff
@@ -0,0 +1,261 @@
...
+- material: publication
+  doi: 10.1016/j.nima.2019.162728
+journal_volume: '958'
+journal_issue: '3'
+is_conference_paper: true
+publication_date: '2020-04-01'
+collaborations: ['NA61/SHINE, CBM and BM@N collaborations']
+documents:
+- key: j.nima.2019.162728.xml
+  url: http://example.org/j.nima.2019.162728.xml
```
michamos, line 76: where is this coming from?

tests/unit/responses/elsevier/j.nima.2019.162728_expected.yml
```diff
@@ -0,0 +1,261 @@
...
+      langtype="en"><authors><author><surname>Kapishin</surname><given-name>M.</given-name></author></authors></contribution><host><issue><series><title><maintitle>Nucl.
+      Phys.</maintitle></title><volume-nr>A982</volume-nr></series><date>2019</date></issue><pages><first-page>967</first-page><last-page>970</last-page></pages></host></reference><source-text
+      id="afs2">M. Kapishin, Nucl.Phys. A982 (2019) 967-970</source-text></bib-reference>
+    source: Elsevier B.V.
+  reference:
+    publication_info:
+      journal_title: Nucl. Phys.
+      journal_volume: A982
+      year: 2019
+      page_start: '967'
```
michamos, line 102: missing `page_end` from `<last_page>`

tests/unit/responses/elsevier/j.nima.2019.162728_expected.yml
```diff
@@ -0,0 +1,261 @@
...
+      inspire_role: author
+- raw_refs:
+  - schema: Elsevier
+    value: <bib-reference id="b9"><label>[9]</label><other-ref id="sb9"><textref><inter-ref
+      id="interref3" href="http://dpnc.unige.ch/bravar/exp/NA61/Bravar-DRS.pdf" type="simple">http://dpnc.unige.ch/bravar/exp/NA61/Bravar-DRS.pdf</inter-ref>.</textref></other-ref></bib-reference>
+    source: Elsevier B.V.
+  reference:
+    persistent_identifiers:
+    - schema: HDL
+      value: http://dpnc.unige.ch/bravar/exp/NA61/Bravar-DRS.pdf
```
michamos, line 205: this is wrong, it's not a handle (`HDL`). I think you need to exclude stuff starting with `http://` (for unrecognized domains) and add them as URLs instead (to be done in the builder).

tests/unit/responses/elsevier/j.nima.2019.162728_expected.yml
```diff
@@ -0,0 +1,261 @@
...
+- raw_refs:
+  - schema: Elsevier
+    value: <bib-reference id="b10"><label>[10]</label><other-ref id="sb10"><textref>O.
+      Gavrishchuk, et al. <inter-ref id="interref4" role="http://www.elsevier.com/xml/linking-roles/preprint"
+      href="arxiv:1406.1599v2[physics.ins-det]" type="simple">arXiv:1406.1599v2[physics.ins-det]</inter-ref>,
+      2014.</textref></other-ref></bib-reference>
+    source: Elsevier B.V.
+  reference:
+    misc:
+    - arXiv:1406.1599v2[physics.ins-det]
```
michamos, line 216: this should be recognized properly as an arXiv ID. Probably needs improvements in the builder, or alternatively detecting the `role`.

tests/unit/responses/elsevier/j.nima.2019.162787_expected.yml
```diff
@@ -0,0 +1,204 @@
...
+copyright_holder: © 2019 Published by Elsevier B.V.
```
michamos, line 2: This looks weird for the holder.

hepcrawl/parsers/elsevier.py
MJedr: I'll try to make them more precise, but since we use the prism namespace the document type node is not the root one, so in order to not use `//` I should use the conditional xpath containing all the top-level document type (like article, simple-article, book, book-review, simple-book etc.)

tests/unit/responses/elsevier/j.nima.2019.162728_expected.yml
MJedr: That is from test which tests the method
```    
def attach_fulltext_document(self, file_name, url):
        self.builder.add_document(file_name, url, fulltext=True, hidden=True)
```

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,621 @@
...
+COPYRIGHT_MAPPING = {
+    'crown': 'Crown copyright',
+    'free-of-copyright': 'None',
+    'full-transfer': 'Publisher',
+    'joint': 'Publisher and the society',
+    'limited-transfer': 'Publisher and the author',
+    'other': 'Authors',
+    'society': 'The society',
+    'unknown': 'Unknown',
+    'us-gov': ' US government'
```
michamos, line 73: are you sure about this? I don't think it's very useful to have `Unknown` in the record, better not put anything in that case.

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,621 @@
...
+        elif self.root.xpath(book_path):
+            doctype = "book-review"
+        elif self.root.xpath("./exam"):
+            doctype = "exam"
+        return doctype
+
+    @property
+    def artid(self):
+        artid = self.root.xpath(
+            './{}/item-info/aid/text()'.format(self.top_level_document_type)
```
michamos, line 182: not sure I understand this: if there's a `simple-article` node, it becomes `article`?

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,621 @@
...
+            material = 'publication'
+
+        return material
+
+    @property
+    def number_of_pages(self):
+        if self.page_start and self.page_end:
+            return int(self.page_end) - int(self.page_start)
+        if self.page_start and not self.page_end:
+            return 1
```
michamos, line 344: as I said previously, you can remove this, it's handled in the builder

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,621 @@
...
+    def get_reference_iter(self, ref_node):
+        """Extract one reference.
+
+        Args:
+            ref_node(scrapy.selector.Selector): a selector on a single
+                reference, i.e. ``<ref>``.
+
+        Returns:
+            generator[dict]: the parsed reference, as generated by
+                :class:`inspire_schemas.api.ReferenceBuilder`
```
michamos, line 543:
```suggestion
        Yields:
            dict: the parsed reference, as generated by
                :class:`inspire_schemas.api.ReferenceBuilder`
```

hepcrawl/parsers/elsevier.py
MJedr: That's whats written in [The Elsevier DTD 5 Family of XML DTDs](https://www.elsevier.com/__data/assets/pdf_file/0010/851626/tagbytag5-v1.9.7.pdf):
```
• crown is used when the author claims Crown copyright. 
• free-of-copyright: this value is used when the item has no copyright, for example in case of an index. 
• full-transfer: this value is used when a full transfer to one of the publisher’s
companies has been received.
• joint is used when a full transfer has been received for an article in a journal whose
copyright is owned jointly by one of the publisher’s companies and a society. 
• limited-transfer is used when the author has granted only limited rights; special
care must be taken for its production.
• no-transfer is used when there is an unresolvable copyright problem and the article may not be published (in principle, documents with this copyright status cannot
occur). This should not be confused with copyright status 001, when copyright has
not yet been transferred. Within ce:figure, ce:textbox or ce:e-component it
is used to signal that the object may not appear online.
• other is used when copyright owner is different from the journal’s copyright owner,
e.g. the authors or their employing institutions. This copyright type is also used in
the ce:copyright within a ce:figure, ce:textbox or ce:e-component. 
• society is used when a full transfer has been received for an article in a journal
whose copyright is owned by a society. 
• unknown: this value is used when the article may be published but the actual status
is unknown. This is, for instance, the case when the copyright transfer form has not
yet been received from the author. 
• us-gov is used when the author is a US government employee and will not transfer
copyright. 
```
To me it looks the most reasonable for the copyright owner in case when no other information is provided.

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+            'adv': 'advertisement',
+            'ann': 'announcement',
+            'brv': 'book-review',
+            'cal': 'calendar',
+            'chp': 'chapter in a book',
+            'cnf': 'conference',
+            'con': 'contents list',
+            'cop': 'copyright information',
+            'cor': 'correspondence',
+            'crp': '',
```
michamos, line 38: ?

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+            'rev': 'review',
+            'sco': 'short communication',
+            'ssu': 'short survey',
+        }
+
+COPYRIGHT_MAPPING = {
+    'crown': 'Crown copyright',
+    'free-of-copyright': 'None',
+    'full-transfer': 'Publisher',
+    'joint': 'Publisher and the society',
```
michamos, line 68:
```suggestion
    'joint': 'Publisher and scientific society',
```

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+    'limited-transfer': 'Publisher and the author',
```
michamos, line 69:
```suggestion
    'limited-transfer': 'Authors and publisher',
```

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+    'other': 'Authors',
+    'society': 'The society',
```
michamos, line 71:
```suggestion
    'society': 'Scientific society',
```

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+    }
+
+    @property
+    def abstract(self):
+        abstract_nodes = self.root.xpath(".//head/abstract/abstract-sec/simple-para")
+
+        if not abstract_nodes:
+            return
+
+        abstract = remove_tags(abstract_nodes[0], **self.remove_tags_config_abstract).strip('/ /n')
```
michamos, line 147: Should the `/`s be `\`s in `strip`?

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+            'prp': 'personal report',
+            'prv': 'product review',
+            'pub': 'publisher\'s Note',
+            'rem': 'removal',
+            'req': 'request for assistance',
+            'ret': 'retraction',
+            'rev': 'review',
+            'sco': 'short communication',
+            'ssu': 'short survey',
+        }
```
michamos, line 62: where are these coming from? it seems to be different from https://www.elsevier.com/__data/assets/pdf_file/0010/784522/ja5_tagbytag5-v1.9.6.3.pdf page 52. Please use the list of abbreviations and names there.

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+    def top_level_document_type(self):
+        possible_article_types = 'self::article or self::simple-article'
+        possible_book_types = 'self::book or self::book-chapter or self::book-review or self::simple-book'
+        possible_top_level_document_types = "./*[{0} or {1} or self::exam]".format(
+            possible_article_types, possible_book_types
+        )
+        top_level_document_type_node = self.root.xpath(possible_top_level_document_types)
+        if top_level_document_type_node:
+            top_level_document_type = top_level_document_type_node[0].xpath('name()').extract_first()
+            return top_level_document_type
```
michamos, line 178: I don't think you have to do this. If you know that the node selectors are at the n-th level, using something like `./*/foo/bar` is perfectly fine. As long as you don't start all your selectors with `//`, it's ok.

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+    @property
+    def article_type(self):
+        """Return a doctype mapped from abbreviation."""
+        abbrv_doctype = self.root.xpath(".//@docsubtype").extract()
+        doctype = ''
+        if abbrv_doctype:
+            doctype = DOCTYPE_MAPPING[get_first(abbrv_doctype)]
+        elif self.top_level_document_type in ['article', 'simple-article']:
+            doctype = 'article'
+        elif self.top_level_document_type in ['book', 'book-chapter', 'book-review', 'simple-book']:
```
michamos, line 159: a book review is not a book, but an article.

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+    @property
+    def journal_volume(self):
+        journal_volume = self.root.xpath("./RDF/Description/volume/text()").extract_first()
+
+        return journal_volume
+
+    @property
+    def keywords(self):
+        keywords = self.root.xpath(
+            "./{}/head/keywords[not(@abr)]/keyword/text/text()".format(self.top_level_document_type)
```
michamos, line 300: why the `[not(@abr)]`. What do those represent?

hepcrawl/parsers/elsevier.py
```diff
@@ -0,0 +1,615 @@
...
+        return license_url
+
+    @property
+    def material(self):
+        if self.article_type in ('erratum', 'addendum'):
+            material = self.article_type
+        else:
+            material = 'publication'
+
+        return material
```
michamos, line 338: I think you can map more:
* `retraction` -> `erratum`
* `removal`-> `erratum`
* `duplicate` -> `erratum`
* `editorial` -> `editorial note`
* `publisher's note` -> `editorial note`

michamos (commented):
I'm a bit confused by all the things happening around `document_type`, `top_level_document_type`, etc. let's discuss it IRL.

hepcrawl/parsers/elsevier.py
MJedr: These are abbreviations used in article, I don't think that they should be included

michamos (approved):

PR #284, opened 2020-09-07
Add s3 to DESY spider instead of FTP

pazembrz:
Currently all additional files for record are intentionally ignored. (After discussion with Micha)

ref:inspirehep/inspirehep#1245


hepcrawl/spiders/desy_spider.py
```diff
@@ -78,27 +80,81 @@ class DesySpider(StatefulSpider):
...
+            # Create resource or client
+            if service in session.get_available_resources():
+                connections.update({service: session.resource(service, **kwargs)})
+            else:
+                connections.update({service: session.client(service, **kwargs)})
+        except UnknownServiceError as exc:
+            self.logger.error("Cannot connect to s3.", exc)
+            raise
+        return connections
+    def get_s3_url_for_key(self, key, bucket=None, expire=86400):
```
michamos, line 143: missing newline

hepcrawl/spiders/desy_spider.py
```diff
@@ -121,62 +177,25 @@ def crawl_local_directory(self):
...
-        self.logger.info('Visited url {}'.format(response.url))
-        file_path = response.body
-        yield Request(
-            'file://{0}'.format(file_path),
-            meta={'source_folder': file_path},
-            callback=self.parse,
-        )
+    def crawl_s3_bucket(self):
+        input_bucket = self.s3_resource.Bucket(self.s3_input_bucket)
+        existing_files = os.listdir(self.destination_folder)
```
michamos, line 182: I would forget about the local files handling, it's not needed if you move the source files to a different bucket.

hepcrawl/spiders/desy_spider.py
```diff
@@ -238,13 +249,20 @@ def parse(self, response):
...
-            hostname=hostname,
-            url_schema=url_schema,
-            ftp_params=ftp_params,
             url=response.url
         )
         self.logger.info('Got %d hep records' % len(parsed_items))
 
+        if self.s3_enabled:
+            s3_file = response.meta['s3_file']
+            self.logger.info("Moving {file} to {processed} bucket and deleting {file} from {incoming} bucket.".format(
```
michamos, line 258:
```suggestion
            self.logger.info("Moving {file} from {incoming} bucket to {processed} bucket.".format(
```

hepcrawl/spiders/desy_spider.py
pazembrz: I would prefer to leave `local` part of spider intact, as it's not part of this task.

hepcrawl/spiders/desy_spider.py
pazembrz: :+1:

hepcrawl/spiders/desy_spider.py
pazembrz: :+1:

hepcrawl/spiders/desy_spider.py
michamos: actually it looks like this method is not used at all. Could you check?

pazembrz: Yeah, you are right. It's leftover from processing additional files I did at first.

michamos (approved):

PR #285, opened 2020-09-18
spiders: add elsevier spider

MJedr:
Ref: inspirehep/inspirehep#1212


docker-compose.test.py2.yml
```diff
@@ -196,12 +203,13 @@ services:
   localstack:
     image: localstack/localstack:latest
     ports:
-      - '4572:4572'
-      - '4566:4566'
+      - "4566-4599:4566-4599"
```
pazembrz, line 206: Why you are exposing all ports? From what I remember s3 is on port 4572 only.

pazembrz: As I can see you are using port 4566 as 4572 is actually deprecated. Then leave only 4566 open.

MJedr: Yes, exactly, I was looking for the reference because I've read about it too ;)

hepcrawl/parsers/elsevier.py
```diff
@@ -341,11 +342,17 @@ def page_end(self):
...
         publication_date_string = self.root.xpath(
             './RDF/Description/coverDisplayDate/text()'
         ).extract_first()
         if publication_date_string:
-            publication_date = PartialDate.parse(publication_date_string)
+            try:
+                publication_date = PartialDate.parse(publication_date_string)
+            except:
+                # in case when date contains month range, eg. July-September 2020
+                publication_date = re.sub("[A-aZ-z]*-(?=[A-aZ-z])", "", publication_date_string)
```
michamos, line 354: so `July-September 2020` becomes 2020? that's not really great.

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,188 @@
+import glob
+import os
+import shutil
+import tempfile
+import xml.etree.ElementTree as et
+import zipfile
+
+import boto3
+import requests
```
michamos, line 9: this is wrong, you shouldn't have to use `requests`. The way `scrapy` works is that you yield a `Request`, which then gets consumed by the `scrapy` middleware, which downloads the response and passes it to the callback, see https://github.com/inspirehep/hepcrawl/blob/1d6c9c7b8716d5b98b5f604828a809b8f04b6e9e/hepcrawl/spiders/aps_spider.py#L104-L107 for an example. This allows `scrapy` to do all the scheduling of requests to allow doing requests in parallel, avoid overloading the server, etc.

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def _get_keys_names_from_bucket(self):
+        keys = set([key.key for key in self.s3_packages_bucket_conn.objects.all()])
+        return keys
+
+    def _download_elsevier_metadata(self):
+        elsevier_batch_download_url = (
+            "https://consyn.elsevier.com/batch/atom?key=" + self.elsevier_consyn_key
+        )
+        packages_metadata = requests.get(elsevier_batch_download_url)
```
michamos, line 98: still using `requests` here

hepcrawl/parsers/elsevier.py
MJedr: No, it leaves the second month from the range.

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,235 @@
...
+import boto3
+import scrapy
+from scrapy import Request, Spider
+
+from . import StatefulSpider
+from ..parsers import ElsevierParser
+from ..utils import ParsedItem, strict_kwargs
+
+
+class ElsevierSpider(StatefulSpider):
```
michamos, line 18: does it need to be a `StatefulSpider`? if you're checking manually which zip files we've already received, I'm not sure there's much point to it.

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,235 @@
...
+            dict(name: url): dict of zip packages names and urls
+
+        """
+        packages_metadata_parsed = et.fromstring(elsevier_metadata)
+        urls_for_packages = {}
+        for children in packages_metadata_parsed.getchildren():
+            if "entry" in children.tag:
+                file_data = children.getchildren()
+                link = file_data[1].attrib["href"]
+                urls_for_packages[file_data[0].text] = link
```
michamos, line 108: maybe it's easier to use a scrapy `Selector` here rather than bare `ElementTree`? (and in any case, you should use the `lxml` version, not the pure python standard library implementation which is very slow).

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,235 @@
...
+        self.new_packages = set(packages_not_in_bucket.keys())
+        return packages_not_in_bucket
+
+    def start_requests(self):
+        elsevier_batch_download_url = self.elsevier_consyn_url
+        yield Request(
+            elsevier_batch_download_url, callback=self.get_packages_from_elsevier
+        )
+
+    def get_packages_from_elsevier(self, response):
```
michamos, line 133:
```suggestion
    def get_packages_from_consyn_feed(self, response):
```
the class and the file are called `elsevier`, no need to repeat it all the time, better use more descriptive names.

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,235 @@
...
+    def create_presigned_url(self, bucket, file, method):
+        url = self.s3_client.generate_presigned_url(
+            ClientMethod=method,
+            Params={"Bucket": bucket, "Key": file},
+            ExpiresIn=920000,
+        )
+        return url
+
+    def _get_keys_names_from_bucket(self):
+        keys = set([key.key for key in self.s3_packages_bucket_conn.objects.all()])
```
michamos, line 91: this will become very slow as time goes on. Maybe it's better to check whether each file exists by doing a request for each zip file in the feed?

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class ElsevierSpider(StatefulSpider):
+    name = "elsevier"
+    start_urls = []
+
+    @strict_kwargs
+    def __init__(
+        self,
+        acces_key_id,
```
michamos, line 25:
```suggestion
        access_key_id,
```

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,235 @@
...
+        Downloads zip packages from s3 and passes to unziping function.
+        """
+        get_url = self.create_presigned_url(
+            bucket=self.packages_bucket_name,
+            method="get_object",
+            file=response.meta["name"],
+        )
+        yield Request(
+            get_url,
+            callback=self.unzip_zip_package_to_s3,
```
michamos, line 179: do you need to download it again? we had it already to populate the bucket with zip file, can't you reuse that instead?

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,235 @@
...
+            meta={"name": response.meta["name"]},
+        )
+
+    def unzip_zip_package_to_s3(self, response):
+        """
+        Extracts the files from zip folders downloaded from elsevier and
+        uploads them with a correct name (article doi) to the correct s3 bucket.
+        """
+        tempdir = tempfile.mkdtemp()
```
michamos, line 188: you should use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory (or a backport if not available on Python 2.7) which can be used as a context manager. Your code has a bug, it's not cleaning up if there are exceptions.

michamos (commented):
not finished yet, will continue tomorrow, but here are already some comments

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,206 @@
...
+                    yield Request(
+                        url,
+                        method="PUT",
+                        body=elsevier_xml,
+                        meta={"name": "{file_doi}.xml".format(file_doi=file_doi),
+                              "data": elsevier_xml},
+                        callback=self.parse_record_from_s3,
+                    )
+
+    def parse_items_from_s3(self, response):
```
michamos, line 183: is this method ever used?

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,206 @@
...
+
+        return ParsedItem(record=parser.parse(), record_format="hep")
+
+    def parse_record(self, record):
+        """Parse an elsevier XML exported file into a HEP record."""
+        with open(record, "r") as f:
+            elsevier_record = f.read()
+
+        parser = ElsevierParser(elsevier_record)
+        return ParsedItem(record=parser.parse(), record_format="hep")
```
michamos, line 206: Is this one used?

What we would need is a way to reharvest a given record, but we need to be able to provide it as an input to scrapy somehow. I think you can leave it out of the current PR, as it's already taken a lot of time and it's not super-urgent, and create an issue for this so we don't forget it.

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,206 @@
...
+                        file="{file_doi}.xml".format(file_doi=file_doi),
+                    )
+
+                    yield Request(
+                        url,
+                        method="PUT",
+                        body=elsevier_xml,
+                        meta={"name": "{file_doi}.xml".format(file_doi=file_doi),
+                              "data": elsevier_xml},
+                        callback=self.parse_record_from_s3,
```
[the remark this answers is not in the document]
MJedr: Here I put the articles in the bucket `inspire-publishers-elsevier-articles` and as a callback I send it to the parser (without downloading it once again)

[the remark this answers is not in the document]
michamos: right, I missed that, sorry

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,206 @@
...
+        Downloads xml files from S3 and passes to parser.
+        """
+        download_url = self.create_presigned_url(
+            bucket=self.files_bucket_name,
+            method="get_object",
+            file=response.meta["name"],
+        )
+        yield Request(download_url, callback=self.parse_record_from_s3)
+
+    def parse_record_from_s3(self, response):
```
michamos, line 194: bu then this function is a misnomer, it's not actually parsing the record from S3 but from the local data.

hepcrawl/spiders/elsevier.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def extract_packages_from_consyn_feed(self, response):
+        """
+        Parse batch feed file from elsevier and downloads new zip packages from Elsevier server.
+        """
+        elsevier_metadata = response.body
+        packages_from_consyn_feed = self._get_package_urls_from_elsevier(elsevier_metadata)
+        for name, url in packages_from_consyn_feed.items():
+            presigned_url = self.create_presigned_url(
+                method="get_object", bucket=self.packages_bucket_name, file=name
```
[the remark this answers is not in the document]
MJedr: OK, but in this scenario I have to handle it with `errback` because Scrapy enable running callback only when response status is 200.

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -1,1046 +1,195 @@
...
-            if not book_title:
-                book_title = ref.xpath(
-                    ".//sb:edited-book/sb:title/ce:inter-ref/text()"
-                ).extract_first()
-        else:
-            book_title = ref.xpath(
-                ".//sb:book//sb:maintitle/text()").extract_first()
-        return book_title
+    def package_already_exists(self, response):
+        pass
```
michamos, line 128: maybe add some logging here?

michamos (approved):
some minor comments remaining, but looks very good, great job! :+1:

PR #288, opened 2020-10-14
fix CDS spider

pazembrz:
FIX: from_exception() takes exactly 5 arguments (4 given)

### DIFF
--- a/hepcrawl/spiders/cds_spider.py
+++ b/hepcrawl/spiders/cds_spider.py
@@ -138,5 +138,6 @@ def _parsed_item_from_marcxml(
                 record_format='hep',
                 exception=repr(e),
                 traceback=tb,
-                source_data=marcxml_record
+                source_data=marcxml_record,
+                file_name=None
             )


hepcrawl/spiders/cds_spider.py
```diff
@@ -138,5 +138,6 @@ def _parsed_item_from_marcxml(
                 record_format='hep',
                 exception=repr(e),
                 traceback=tb,
-                source_data=marcxml_record
+                source_data=marcxml_record,
+                file_name="__missing__"
```
drjova, line 142: `None` doesn't work?

pazembrz: file_name on ParsedItem is used in few places. It's safer to set something in there.

drjova: What if it tries to retrieve from somewhere, isn't it safer just to use None?

pazembrz: It's not retrieving it from anywhere, as it's a filename extracted from url (which is not available in this context) so it cannot be accessed locally on spider. 
I don't like it as I'm not sure what might happen in this spider when file_name will be None. In few places we are processing it as it's a string, not expecting it to be None so probably it will throw more exceptions when it will be None.

drjova: let's just put `''` then, I see it's been used in few places. At least if there is a `if file_name
` it will be `False`

pazembrz: I checked few more places in code. I will try with None. I don't like it but it will be cleaner than empty string IMHO.

drjova (approved):

PR #289, opened 2020-10-19
elsevier: restrict records harvested from elsevier

MJedr:
Ref: inspirehep/inspirehep#1446


michamos (approved):
minor comment, but looks good :+1:

docker-compose.test.py2.yml
```diff
@@ -199,7 +199,7 @@ services:
       test: "redis-cli -h 127.0.0.1 ping| grep PONG"
 
   localstack:
-    image: localstack/localstack:latest
+    image: localstack/localstack:0.11.6
```
drjova, line 202: 👍

docker-compose.test.py2.yml
```diff
@@ -213,7 +213,7 @@ services:
       timeout: 5s
       interval: 5s
       retries: 5
-      test: "curl -k localhost:4566|grep running"
+      test: "curl -k localstack:4566|grep running"
```
drjova, line 216: do you we need to change to `localstack`?

MJedr: We set the `HOSTNAME_EXTERNAL=localstack` and `HOSTNAME=localstack` so I think that we should use localstack here (locally when I test with `docker-compose` the log is `Unable to find forwarding rule for host "localhost:4566"`)

PR #291, opened 2020-10-23
elsevier: fix url for attaching documents

MJedr:


michamos (approved):

michamos:
I'm curious: how was this not caught by tests?

MJedr:
The test seems to check only if the attached documents have the S3 url and file name url included and they have them. The problem starts when the pipeline things happen with GET request from the url to the attached document to this new bucket.

PR #292, opened 2020-10-23
Fix APS spider - document url header

pazembrz:
APS XML document requires `Accept: text/xml` header to properly download XML file.

### DIFF
--- a/hepcrawl/pipelines.py
+++ b/hepcrawl/pipelines.py
@@ -56,6 +56,7 @@ class DocumentsPipeline(FilesPipeline):
         )
 
     def get_media_requests(self, item, info):
+        LOGGER.info("item: :", item)
         if item.get('file_urls'):
             LOGGER.info(
                 'Got the following files to download:\n%s', pprint.pformat(
@@ -63,9 +64,9 @@ class DocumentsPipeline(FilesPipeline):
                 )
             )
             return [Request(x) for x in item.get(self.files_urls_field, [])]
-        return list()
+        return item.get("file_requests", [])
 
-    def generate_presigned_s3_url(self, path, expire=86400):
+    def generate_presigned_s3_url(self, path, expire=7776000):
         bucket_location = get_project_settings().get("DOWNLOAD_BUCKET", "documents")
         LOGGER.info("Generating presigned url for: %s in %s", path, bucket_location)
         return self.store.s3_client.generate_presigned_url(
--- a/hepcrawl/spiders/aps_spider.py
+++ b/hepcrawl/spiders/aps_spider.py
@@ -123,11 +123,11 @@ class APSSpider(LastRunStoreSpider):
         file_name = self._file_name_from_url(response.url)
         parser.attach_fulltext_document(file_name, response.url)
         record = parser.parse()
-        file_urls = [document['url'] for document in record.get('documents', [])]
+        file_requests = [Request(url=document['url'], headers={'Accept': 'text/xml'}) for document in record.get('documents', [])]
         return ParsedItem(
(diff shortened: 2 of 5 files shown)


hepcrawl/pipelines.py
```diff
@@ -62,10 +62,10 @@ def get_media_requests(self, item, info):
                     item['file_urls']
                 )
             )
-            return [Request(x) for x in item.get(self.files_urls_field, [])]
+            return [Request(**x) for x in item.get(self.files_urls_field, [])]
         return list()
 
-    def generate_presigned_s3_url(self, path, expire=86400):
+    def generate_presigned_s3_url(self, path, expire=7776000):
```
michamos, line 68: why?

hepcrawl/spiders/aps_spider.py
```diff
@@ -123,7 +123,7 @@ def _parse_jats(self, response):
         file_name = self._file_name_from_url(response.url)
         parser.attach_fulltext_document(file_name, response.url)
         record = parser.parse()
-        file_urls = [document['url'] for document in record.get('documents', [])]
+        file_urls = [{"url": document['url'], "headers": {'Accept': 'text/xml'}} for document in record.get('documents', [])]
         return ParsedItem(
```
michamos, line 127: `file_urls` doesn't take URLs any longer, but a list of `dict`s now, so the name doesn't match. Maybe we should change the API to take a list of `Request`s instead?

hepcrawl/pipelines.py
pazembrz: 86400 seconds is only 24 hours. If workflow will be restarted next day because of some problem then url with 86400 seconds won't be valid anymore.

hepcrawl/spiders/aps_spider.py
pazembrz: I don't understand. In results scrapy is producing list of tuples with url in there:
`2020-10-23 09:24:02 [hepcrawl.pipelines] INFO: results: [(True, {'url': 'http://aps-http-server.local/PhysRevD.96.095036.xml', 'path': 'full/b99616c5061a542667fb4fa1d5a8ab750a15c731.xml', 'checksum': '4d8a21a5aeb643400328ed1915e70f66'})]`
And then we are matching this url with url in record. Looks lie it still works. there are even tests to check if document url in record is what we want. And it's still passing.

michamos: I'm not saying it's broken, I'm saying you're making it more complex by:
* having content of `file_urls` not match the name (not a list of URLs any longer)
* needing to pass a magic dict, which turns out to be `kwargs` to `Request`

My suggestion:
* pass a list of `Request`s to `ParsedItem` instead of only its args
* replace the `ParsedItem` `file_urls` parameter by something more explicit, like `document_requests`

michamos (approved):

hepcrawl/pipelines.py
```diff
@@ -56,16 +56,17 @@ def __init__(self, store_uri, *args, **kwargs):
...
+        LOGGER.info("item: :", item)
         if item.get('file_urls'):
             LOGGER.info(
                 'Got the following files to download:\n%s', pprint.pformat(
                     item['file_urls']
                 )
             )
             return [Request(x) for x in item.get(self.files_urls_field, [])]
-        return list()
+        return item.get("file_requests", list())
```
michamos, line 67: Minor: 
```suggestion
        return item.get("file_requests", [])
```

pazembrz: :+1:

PR #293, opened 2020-10-26
Fix DESY S3 support

pazembrz:
* Fix error when processing desy document without attached documents
* Fix missing external documents when local and external documents mix was added to main document

### DIFF
--- a/hepcrawl/spiders/desy_spider.py
+++ b/hepcrawl/spiders/desy_spider.py
@@ -126,7 +126,7 @@ class DesySpider(StatefulSpider):
     def s3_url(self, s3_file=None, expire=86400):
         return self.s3_url_for_file(s3_file.key, bucket=s3_file.bucket_name, expire=expire)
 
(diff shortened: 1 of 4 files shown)


hepcrawl/spiders/desy_spider.py
```diff
@@ -249,13 +249,14 @@ def _parsed_items_from_marcxml(
                     files_to_download = []
                     self.logger.info("Parsed document: %s", parsed_item.record)
                     self.logger.info("Record have documents: %s", "documents" in parsed_item.record)
-                    for document in parsed_item.record.get('documents'):
+                    for document in parsed_item.record.get('documents', []):
```
drjova, line 252:
```python
new_documents.append(document)
``` 
and you can remove it from `if/else`

pazembrz: :+1:

drjova (approved):

PR #294, opened 2020-10-27
aps parser: handle missing affiliations gracefully

MJedr:
Ref: inspirehep/inspirehep#1465


michamos (commented):
did you edit one of the existing test cases to "fake" the isuse? I think it would be better to use a real record so we keep track of real problems.

MJedr:
Yes, I've changed the reference id assigned to one of the authors for not existing one. If that's important to keep the original records I can create the mock one for tests or just leave tests as they are.

michamos:
you could use 10.1103/PhysRevD.102.014505 which exhibits this issue.

michamos (approved):

PR #296, opened 2021-01-26
arxiv parser: split multiple doi

MJedr:
Ref: inspirehep/inspirehep#1650

### DIFF
--- a/hepcrawl/parsers/arxiv.py
+++ b/hepcrawl/parsers/arxiv.py
@@ -11,6 +11,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+from itertools import chain
 import re
 
 import six
@@ -35,6 +36,8 @@ RE_THESIS = re.compile(
     re.I | re.U,
 )
 RE_PAGES = re.compile(r'(?i)(\d+)\s*pages?\b')
+
+RE_DOIS = re.compile(r'[,;\s]+(?=\s*10[.]\d{4,})')
 
 
 class ArxivParser(object):
@@ -227,8 +230,9 @@ class ArxivParser(object):
     @property
     def dois(self):
         doi_values = self.root.xpath('.//doi/text()').extract()
+        doi_values_splitted = chain.from_iterable([re.split(RE_DOIS, doi) for doi in doi_values])
         dois = [
-            {'doi': value, 'material': 'publication'} for value in doi_values
+            {'doi': value, 'material': 'publication'} for value in doi_values_splitted
         ]
 
         return dois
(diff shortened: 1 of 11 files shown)


hepcrawl/parsers/arxiv.py
```diff
@@ -36,6 +37,8 @@
 )
 RE_PAGES = re.compile(r'(?i)(\d+)\s*pages?\b')
 
+RE_DOIS = re.compile(r'[,;\s](?=\s*)')
```
tsgit, line 40: apologies for jumping in here, I have 2 comments: 
the name `RE_DOI` is misleading, since it doesn't match a DOI, instead it is supposed to match a `separator` between 2 DOIs
The regexp should be refined to specifically split between 2 DOIs and not at any `,` or `;`  I suggest `r'[,;\s]+(?=\s*10[.]\d{4,})'` 
```
In [1]: dois = ['10.7916/d8-bbr3-zv47,10.1016/j.nuclphysa.2021.122138;10.1016/j.nuclphysa.2020.122133 10.1103/PhysRevLett.119.231101, 10.1142/S0218271817430222']

In [2]: DOI_SEP = r'[,;\s]+(?=\s*10[.]\d{4,})'

In [3]: split_dois = ([re.split(DOI_SEP, doi) for doi in dois])

In [4]: split_dois
Out[4]: 
[['10.7916/d8-bbr3-zv47',
  '10.1016/j.nuclphysa.2021.122138',
  '10.1016/j.nuclphysa.2020.122133',
  '10.1103/PhysRevLett.119.231101',
  '10.1142/S0218271817430222']]
```

MJedr: Thank you, indeed this regex will be more precise.

hepcrawl/testlib/utils.py
```diff
@@ -39,3 +39,8 @@ def deep_sort(element):
         return sorted([deep_sort(item) for item in element])
 
     return element
+
+
+def sort_list_of_records_by_record_title(list_to_sort):
+    """Sort list of records by record title"""
+    return sorted(list_to_sort, key=lambda k: k['titles'][0]['title'])
```
michamos, line 46: why do you need this? how come the order is not deterministic?

michamos (commented):
small question, looks ok for the rest.

hepcrawl/testlib/utils.py
MJedr: Apparently it seems that they come in non-determinicstic order. For example for wsp spider the tests are using a file `wsp_smoke_records.json` in which there are two records - the first one has the titles `{'source': 'WSP', 'title': u'Article-title\u2019s'}` and the second one `[{'source': 'WSP', 'title': u'Article-title L\xe9vy char'}]` but in tests the results comes in reverted order [here is the error](https://travis-ci.org/github/inspirehep/hepcrawl/jobs/756428402). The same happens for other spiders and without ordering the list of results they fail.

michamos (approved):
that's really odd, but it doesn't matter for the correctness of hepcrawl, so let's merge it.

PR #297, opened 2021-02-03
aps spider: add authorization to requests

MJedr:
Ref: inspirehep/inspirehep#1669

### DIFF
--- a/hepcrawl/spiders/aps_spider.py
+++ b/hepcrawl/spiders/aps_spider.py
@@ -52,7 +52,7 @@ class APSSpider(LastRunStoreSpider):
     name = 'APS'
 
     @strict_kwargs
-    def __init__(self, from_date=None, until_date=None,
+    def __init__(self, aps_token, from_date=None, until_date=None,
                  date="published", sets=None, per_page=100, aps_url="http://harvest.aps.org/v2/journals/articles",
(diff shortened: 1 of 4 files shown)


hepcrawl/spiders/aps_spider.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, from_date=None, until_date=None,
         self.until_date = until_date
         self.per_page = per_page
         self.date = date
+        self.aps_token = aps_token
```
michamos, line 66: maybe store the authorization header instead, as that's what you're using everywhere?
```suggestion
        self.auth_header = "Bearer {}".format(aps_token)
```

michamos (approved):
small suggestion, but looks ok

hepcrawl/spiders/aps_spider.py
```diff
@@ -85,7 +86,7 @@ def url(self):
     def start_requests(self):
         """Just yield the url."""
         started_at = datetime.utcnow()
-        yield Request(self.url)
+        yield Request(self.url, headers={self.auth_header})
```
michamos, line 89: typo? this should be a `dict`

MJedr: Thank you!

PR #309, opened 2021-08-10
logging: File log handler changed to Stream handler. Sentry extension added.

oguzdemirbasci:


setup.py
```diff
@@ -74,6 +74,7 @@
     'sentry': [
         'raven~=6.0,>=6.2.1',
         'scrapy-sentry',
+        'sentry-sdk==1.3.0',
```
michamos, line 77: also included on L44. Why is this added twice? BTW, do we still need `raven` and `scrapy-sentry` now that you've changed the sentry client?

michamos: I think you can also remove https://github.com/inspirehep/hepcrawl/blob/7cd64d4887b38127d8a0755423929b32b7da4cfa/hepcrawl/settings.py#L110

oguzdemirbasci: `scrapy-sentry`s readme file says it doesn't report the errors in scrapy pipelines.

michamos: but you've added a new extension using `sentry-sdk`, which is precisely why you can remove `scrapy-sentry` and `raven` now.

hepcrawl/settings.py
```diff
@@ -107,8 +107,7 @@
 if SENTRY_DSN:
     EXTENSIONS = {
         'hepcrawl.extensions.SentryLogging': 100,
-        'scrapy_sentry.extensions.Errors': 200,
-        'hepcrawl.extensions.ErrorHandler': 300,
+        'hepcrawl.extensions.ErrorHandler': 200,
```
drjova, line 110: shouldn't be `300`?

drjova (approved):

PR #314, opened 2021-10-01
desy: fix bug with encoding

drjova:
* Ref: inspirehep/inspirehep#2227

### DIFF
--- a/hepcrawl/spiders/desy_spider.py
+++ b/hepcrawl/spiders/desy_spider.py
@@ -225,7 +225,7 @@ class DesySpider(StatefulSpider):
             if not list_items:
                 list_items = root.findall('.//record')
 
-        return [etree.tostring(item) for item in list_items]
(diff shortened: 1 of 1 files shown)


hepcrawl/spiders/desy_spider.py
```diff
@@ -225,7 +225,7 @@ def _get_marcxml_records(self, response_body):
             if not list_items:
                 list_items = root.findall('.//record')
 
-        return [etree.tostring(item) for item in list_items]
+        return [etree.tostring(item, encoding='utf8') for item in list_items]
```
michamos, line 228: I would use `utf-8` here or `UTF-8`. The `utf8` one is the one which was buggy in the OAI server, so we might get surprises there.

drjova:
@michamos updated

michamos (approved):

PR #324, opened 2022-11-11
aps parser: `Physics` journal should be translated to `APS Physcis`

MJedr:
cern-sis/issues-inspire#170


drjova:
@michamos do we have to run a script to fix the existing ones?

drjova (approved):

PR #325, opened 2023-01-19
use imprints date from tag rather than publication_date

MJedr:
ref: cern-sis/issues-inspire#206


hepcrawl/parsers/elsevier.py
```diff
@@ -372,6 +372,16 @@ def page_end(self):
...
 
+    @property
+    def imprints_date(self):
+        imprints_date = self.root.xpath(
+            "string(./RDF/Description/availableOnlineInformation/availableOnline)"
+        ).extract_first()
+        if imprints_date:
+            return PartialDate.parse(imprints_date).dumps()
+        if self.publication_date:
+            return self.publication_date.dumps()
```
drjova, line 383: it's ok, I would just move the logic with the `publication_date` to the `parse` since `imprints_date` should handle only one thing

MJedr: Yes, but with a cost of a lower readability imo. But I can move it there.

hepcrawl/parsers/elsevier.py
```diff
@@ -128,7 +128,9 @@ def parse(self):
         for keyword in self.keywords:
             self.builder.add_keyword(keyword)
         self.builder.add_imprint_date(
-            self.publication_date.dumps() if self.publication_date else None
+            self.imprints_date if self.imprints_date else (
+                self.publication_date.dumps() if self.publication_date else None
+            )
```
drjova, line 133: could we simplify this? 😅

drjova (approved):

PR #358, opened 2025-05-15
parser: parse from simple-head too

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/429


ioannistsanaktsidis:
A note related to this PR. The repo is not python 3 compatible, meaning we can build it but tests are not running etc... Yet we release it with python3  -> https://github.com/inspirehep/hepcrawl/pull/358/files#diff-20e0e050358c0425896e7b9edb659fec4ed949bb350a35841c0d616e1971cc6bR22. What was the reason for that? Was it just to release it for now as we using it only in python2(inspire-next) and fix it in the future? cc @drjova

ioannistsanaktsidis:
Also, usecase -> https://github.com/inspirehep/hepcrawl/pull/358/files#diff-c9b75f1849ee2f819f04cabf4463702f66d1c748fb8ef795ea2a3ea81d06309eR165. Fails on CI, succeeds locally with exactly the same setup. Issue seems to be here https://github.com/inspirehep/hepcrawl/blob/master/hepcrawl/spiders/pos_spider.py#L112-L117. For some reason it is not extracted correctly on the CI. To be checked further

michamos:
Not directly related to `simple-head`, by I noticed two additional issues:
* ORCIDs don't seem to be extracted
* In the example file, the DOI of the original paper is mentioned in
```xml
<ce:document-thread><ce:refers-to-document id="rd0010"><ce:pii>S0550-3213(22)00251-6</ce:pii><ce:doi>10.1016/j.nuclphysb.2022.115900</ce:doi></ce:refers-to-document></ce:document-thread>
```
It would be good to extract this DOI too so the erratum gets matched with the original publication, otherwise it will create new records in INSPIRE which we don't want. So in this case, DOIs should look like
```json
{
    "dois": [
        {"material": "erratum", "value": "10.1016/j.nuclphysb.2022.115991"},
        {"material": "publication", "value": "10.1016/j.nuclphysb.2022.115900"}
    ]
}
```
(order doesn't matter)

ioannistsanaktsidis:
Thanks @michamos , will create another ticket to handle these findings.

michamos:
OK, but note that this should not go to prod without the related DOI extraction, as it will cause a mess.

ioannistsanaktsidis:
But isn’t that the case already ? Don’t we crawl elsevier papers like that one ?

tests/unit/test_parsers_elsevier.py
```diff
@@ -163,3 +163,34 @@ def test_imprints_date_should_be_taken_from_avaliable_online():
...
+    parser = get_parser_by_file("j.nuclphysb.2022.115991.xml")
+    result = parser.parse()
+    expected_title = six.ensure_text('Erratum to \u201cThe fifth-order post-Newtonian Hamiltonian dynamics of two-body systems from an effective field theory approach\u201d [Nucl. Phys. B 983 (2022) 115900]')
+    assert result['titles'][0]['title']  == expected_title
+
+
+def test_authors_should_be_taken_from_simple_head_if_no_head():
+    parser = get_parser_by_file("j.nuclphysb.2022.115991.xml")
+    result = parser.parse()
+    expected_authors = [{'emails': [six.ensure_text('Johannes.Bluemlein@desy.de')], 'full_name': six.ensure_text('Bl\xfcmlein, J.')}, {'full_name': six.ensure_text('Maier, A.')}, {'full_name': six.ensure_text('Marquard, P.')}, {'full_name': six.ensure_text('Sch\xe4fer, G.')}]
```
drjova, line 178: it's missing formatting :)

hepcrawl/parsers/elsevier.py
```diff
@@ -205,6 +211,11 @@ def collaborations(self):
         collaborations = self.root.xpath(
             "./*/head/author-group//collaboration/text/text()"
         ).extract()
+        if not collaborations:
+            collaborations = self.root.xpath(
+                "./*/simple-head/author-group//collaboration/text/text()"
```
drjova, line 216: why the path here is different from abstract?

drjova (changes requested):
Thanks @ioannistsanaktsidis few comment for clarification and formatting

drjova (approved):

PR #362, opened 2025-07-22
add elsevier x-els-api-key header

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/1013

### DIFF
--- a/hepcrawl/spiders/elsevier_spider.py
+++ b/hepcrawl/spiders/elsevier_spider.py
@@ -84,6 +84,7 @@ class ElsevierSpider(StatefulSpider):
         packages_bucket_name,
         files_bucket_name,
         elsevier_authorization_data_base64_encoded,
+        elsevier_api_key,
         elsevier_consyn_url,
         s3_host="https://s3.cern.ch",
         *args,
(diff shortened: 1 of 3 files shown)


ioannistsanaktsidis (approved):
Looks good! Just to double check if passing both headers has an effect

hepcrawl/spiders/elsevier_spider.py
```diff
@@ -130,7 +132,8 @@ def _get_package_urls_from_elsevier(self, elsevier_metadata):
         return urls_for_packages
 
     def start_requests(self):
-        request_headers = {'Authorization': 'Basic ' + self.elsevier_authorization_data_base64_encoded}
+        request_headers = {'Authorization': 'Basic ' + self.elsevier_authorization_data_base64_encoded,
+                            'x-els-api-key': self.elsevier_api_key}
```
drjova, line 136: We could have if else depending on which one we set, this way will allow to switch between them :)

PascalEgn:
@ioannistsanaktsidis @drjova Added the if else statement for the header depending on what is provided. If its fine we can merge and prepare the next bump.

PR #364, opened 2026-06-17
elsevier: use key in all requests

DonHaul:
* ref: cern-sis/issues-inspire/issues/1547


DonHaul:
tests failing because py2 images are no longer supported

drjova (approved):

